PR #4, opened 2017-04-03
BCN-JuanCarlos-Giuseppe

gfrr:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,57 @@
...
+
+console.log(loremipsum.split(" ").length);
+console.log(loremipsum.split(" et ").length - 1);
```
tawebbcn: #checked - well done

PR #9, opened 2017-04-03
BCN-ANDREA-AND-EDUARD

edtrias:
First Exercise

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,125 @@
 // Names and Input
 
+var hacker1 = "James";
+var hacker2 = prompt("What's your name?");
 
 //Conditionals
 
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,125 @@
...
+
+var ipsumByWords;
+
+//count how many words are there in the lorem ipsum
+//First we get rid of the ". and ," symbols, and then we convert to an Arrey via Split()
+
+function separateIpsum() {
+  ipsumByWords = ipsum.replace(/,/g, "").replace(/\./g, "").split(" ");
+  var numberWords = ipsumByWords.length;
```
matilore: you can also concatenate methods, so you can apply the .length to previous line

PR #30, opened 2017-10-16
MIA Alejandro Poggi, Marcos Palacios

ShinCero:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,71 @@
...
+var sPal = "";
+for (var i= s.length -1; i >= 0; i-=1) {
+    sPal +=s[i];
+}
+ if(s === sPal){
+   console.log(true);
+ }else{
+   console.log(false);
+ }
+
```
dsoraluz: This is an interesting way to solve the palindrome exercise.

Another way may would have been to use the split(), reverse(), and join() methods to help you in your comparison.

dsoraluz:
Great job on getting all the past the palindrome exercise.

PR #32, opened 2017-10-16
MIA Chu & Ariel

Chu13:


dsoraluz:
Great job on getting all the way to the Palindrome exercise. 

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,50 @@
...
+  horizontalName2 += hacker2[i];
+}
+ console.log(horizontalName2);
+
+ //Conditionals
+
+ var palindrome = prompt ("Type any word");
+ palindrome = palindrome.replace(/\s/g,"");
+
+ if (palindrome === palindrome.split("").reverse().join("")) {
```
dsoraluz: Great use of the split(), reverse(), and join() functions here.

PR #56, opened 2018-01-23
Mosi River - Miami

mosileko:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+  if (driverLength > navigatorLength){
+    console.log("The Driver has the longest name, it has " + driverLength + " characters");
+  } else if (driverLength < navigatorLength) {
+    console.log("Yo, navigator got the longest name, it has " +  navigatorLength  + "  characters");
+  } else {
+    console.log("wow, you both got equally long names, " + navigatorLength +  "characters!!");
+  }
+}
+
+function driverLettersCaps(){
```
sandrabosk: The assignment here is to print all the characters of the driver's name, separated by a space and in capitals, so we have to do something like: variable.split().join().toUpperCase(). This is just a hint. What would be a solution?

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+  var result = "";
+  for (var i = 0; hacker1.length > i  ;i++ ){
+    result += hacker1[i].toUpperCase();
+    result += " ";
+  }
+  console.log(capSize);
```
sandrabosk: Is capSize defined?

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+}
+
+function printReverse(){
+  result = hacker2.split('').reverse().join("");
```
sandrabosk: .reverse("") is missing the quotes.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+  if (hacker1 < hacker2) {
+    console.log("The driver's name goes first");
+  } else if (hacker1 > hacker2) {
+    console.log("Yo, the navigator goes first definitely");
+  } else {
+    console.log("What?! You both got the same name?");
+  }
+}
+
+
```
sandrabosk: Can you explain the lexicographicOrder() function?

starter-code/basic-algorithms.js
mosileko: Changed the var name halfway through, forgot to edit the function call.

PR #60, opened 2018-01-23
[MIA] [Martin - Genet]

CortezExpress:
this is my practice pull request. i don't remember the code in this but if this is the one where i needed to print the navigator backwards than i need help with that. my code ran, the ".length" method worked with the name i typed into the prompt but then it printed the word navigator () and not the name i typed in


sandrabosk:
Please submit the rest of the exercise and make pull request. Thank you.

PR #61, opened 2018-01-23
[MIA] [Stephen - Jackson]

stephjackson:
issues:
-ignoring characters in palindromes/lorem ipsum probably needs a string method that uses regular expressions but i did it with an if statement

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,80 @@
-// Names and Input
-
+//Names and input
+var hacker1 = "Bob";
+var hacker2 = prompt("What is the navigator's name?");
+console.log("The driver's name is " + hacker1);
+console.log("The navigator's name is " + hacker2);
 
 //Conditionals
+//Measures length of names, outputs string based on who is longer.
+if (hacker1.length > hacker2.length) {
+  console.log("The driver has the longest name, it has " + hacker1.length + " characters");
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,116 @@
-// Names and Input
-
+//Names and input
+var hacker1 = "Bob";
+// var hacker2 = prompt("What is the navigator's name?");
+var hacker2 = "Stephen";
```
sandrabosk: We suppose to do something like this:

var hacker2 = prompt("What's navigator's name?");
console.log("The navigator's name is " + hacker2 + ".");

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,116 @@
...
+if (hacker1.length > hacker2.length) {
+  console.log("The driver has the longest name, it has " + hacker1.length + " characters");
+} else if (hacker2.length > hacker1.length) {
+  console.log("The navigator has the longest name, it has " + hacker2.length + " characters");
+} else {
+  console.log("Wow, you both have equally long names, " + hacker1.length + " characters!")
+}
+
+//Splits driver string into capital letters separated by spaces.
+var capSpace = ""
```
sandrabosk: Great!

sandrabosk:
Perfect! Keep it like this!

PR #72, opened 2018-02-01
MIA Heli

Helieber:
Made some changes!


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,53 @@
...
+    console.log("wow, you both got equally long names, " + hacker1.length + "characters!!")
+  }
+  
+  var string = " "
+  
+  for (var i = 0; i < hacker1.length; i++) {
+   string = string + hacker1[i].toUpperCase() + " "
+  }   
+   console.log(string)
+
```
sandrabosk: // Print all the characters of the navigator's name, in reverse order. ie. "nhoJ"
You're missing this step. Try applying .split, .reverse and .join methods.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,53 @@
...
+   var sentence1 = ["The driver's name goes first", "Yo, the navigator goes first definitely", "What?! You both got the same name?"]
```
sandrabosk: To check if driver's or navigator's name goes first in lexicographic order try applying .localeCompare method.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,53 @@
...
+
+    console.log(sentence1.sort())
+
+// Palindromo
+
+function checkPalindrome(stringCheck){
```
sandrabosk: Very easy way to check if the given word is palindrome is using .split, .reverse, .join methods. Try something like this: 
```
var str = prompt("Let's check the palindrome. Give me a word:");

function checkPalindrome(str) {
  return (
    str ==
    str
      .split("")
      .reverse()
      .join("")
  );
}
var value = checkPalindrome(str);
// console.log("value:",value)

if (value === true) {
  console.log("The word is palindrome.");
} else {
  console.log("The word isn't palindrome.");
}
```

sandrabosk:
Good start Heli. Please check the comments I left here for you. Thank you for submitting.

PR #76, opened 2018-03-19
PAR Pierre & J-B  

PierreJab:


ta-web-paris:
Good job!
However be careful to the semi colons.
For the palindrome exercice, take a look on it if you have time :
https://medium.freecodecamp.org/two-ways-to-check-for-palindromes-in-javascript-64fea8191fd7

PR #79, opened 2018-03-19
PAR Brianna & Leo P.

ycrtw93:


ta-web-paris:
Looks ok to me, if you want to go further and try to resolve the iteration 8, you should check this ethod localeCompare, it may help you !

PR #82, opened 2018-03-19
PAR Luke & Renaud & Elise

lukegeno:
We are not sure about step 8...

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,50 @@
-// Names and Input
+console.log ("I'm Ready!");
+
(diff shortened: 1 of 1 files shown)


ta-web-paris:
You can also use this method : 

var n = hacker1.localeCompare(hacker2)

if (n < 0) { console.log("The driver's  name goes first")}
else if (n > 0){console.log("The navigator's name goes first")}
else {console.log("Wow, you both have the same name!")}


It will compare two string (hacker1 & hacker2) and it will return a number between 1 and -1, then you just have to check if it's greater, equal or lower.

PR #83, opened 2018-03-19
PAR Maggie & Jeremie

JeremieHercberg:


starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,51 @@
...
+var hacker2 = 'Jeremie';
+length2 = hacker2.length;
+var reversename = "";
+for (var i = length2 - 1 ; i >= 0 ; i--){
+  console.log(hacker2[i]);
+  reversename += hacker2[i] ;
+}
+console.log(reversename);
+
+if(hacker1[0] === hacker2[0] ){
```
ta-web-paris: In your case the driver name is always first, you should try this method to compare two string (.localeCompare())
This will return : 
-     -1 if the reference string is sorted before the compareString
-     0 if the two strings are equal
-     1 if the reference string is sorted after the compareString

You can use it like this in your code : 

var n = hacker1.localeCompare(hacker2)

if (n < 0) { console.log("The driver's  name goes first")}
else if (n > 0){console.log("The navigator's name goes first")}
else {console.log("Wow, you both have the same name!")}

PR #94, opened 2018-04-02
MIA-Neil_Dali-Exercise

rojasdali:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,158 @@
...
+
+//Print all the characters of the driver's name, separated by a space and in capitals ie. "J O H N"
+
+var s = "";
+var t = "";
+
+for (i = 0; i < hacker1.length; i++)
+{
+  s += hacker1[i].toUpperCase() + " ";
+
```
sandrabosk: Good approach although the cleaner solution is using JS methods:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .join(" ")
      .toUpperCase()
  );
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,158 @@
...
+}
+console.log(s);
+
+
+//Print all the characters of the navigator's name, in reverse order. ie. "nhoJ"
+
+for (i = hacker2.length-1; i >= 0; i--)
+{
+  t += hacker2[i]
+
```
sandrabosk: The same as above:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .reverse(" ")
      .join("")
  );
}
```

sandrabosk:
Amazing work! Thank you for submitting!

PR #95, opened 2018-04-02
MIA - KatAlderman - Lab-Javascript-Basic-Algorithms

katalderman:
Submitting my work for review!

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,63 @@
 // Names and Input
+var hacker1 = "Kat";
+console.log("The driver's name is " + hacker1);
 
+var hacker2 = prompt("What is your name?");
+console.log("The navigator's name is " + hacker2);
 
 //Conditionals
 
+if (hacker1.length > hacker2.length){
+  
+  console.log("The driver has the longest name, it has " + hacker1.length);
+}
+   else if (hacker2.length > hacker1.length) {
+    
+    console.log("The navigator has the longest name, it has " + hacker2.length);
+  }
+  
+  else {
+    
+    console.log(" you both have equally long names" + hacker1.length );
+  }
+  
+  for (var x=0;x <= hacker1.length; x++) {
+  console.log(hacker1.charAt(x).toUpperCase(x));
+  }
+  
+
+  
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+  }
+  
+  else {
+    
+    console.log(" you both have equally long names" + hacker1.length );
+  }
+  
+  for (var x=0;x <= hacker1.length; x++) {
+  console.log(hacker1.charAt(x).toUpperCase(x));
+  }
```
sandrabosk: Great approach! The other way could be:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .join(" ")
      .toUpperCase()
  );
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+
+  
+  function reverseString(hacker1) {
+    var splitString = hacker1.split("");
+    var reverseArray = splitString.reverse();
+    var joinArray = reverseArray.join("");
+    return joinArray;
+  }
+  console.log(reverseString(hacker1));
+
```
sandrabosk: Again good way of thinking. The same but maybe a bit shorter way would be:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .reverse(" ")
      .join("")
  );
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+  // console.log(array[0]);
+  
+  if (hacker1 < hacker2 ){
+    
+    
+    console.log("The driver's name goes first");
+    
+  }
+  
+  else if (hacker1 > hacker2){
```
sandrabosk: The logic is there but the syntax is not. We could use something like this:
```
var n = hacker1.localeCompare(hacker2);
if (n === 1) {
} else if (n === -1) {
} else {
}
console.log(n);
```
Try googling .localeCompare() to see how this method works.

sandrabosk:
Good job so far. Try finishing the bonus part if you have time. Thank you for submitting.

PR #99, opened 2018-04-02
MIA-Jens Heilemann- JS Basic Algorithms

jkai7:


ianizaguirre:
Hi here is some feedback....

1. For "Loops" you were asked to make 3 conditions but you only made one 🙁 ...
* You wrote:
```js
if (hacker1.length > hacker2.length) {
  console.log("Driver has longer name, it has 7 characters");
}
```
* It should have been:
```js
if (hacker1.length > hacker2.length) {
  console.log('The Driver has the longest name, it has ' + hacker1.length + ' characters');
} else if (hacker1.length < hacker2.length) {
  console.log('Yo, navigator got the longest name, it has ' + hacker2.length + ' characters');
} else {
  console.log('wow, you both got equally long names, ' + hacker1.length + ' characters');
}
```
2. You forgot to do the last part which asks about printing certain strings depending on the lexicographic order (AKA alphabetical order).

Aside from that, good job and keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #100, opened 2018-04-02
Robert Gelbman

RobGelbman:
I worked with Randy and Daniel

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,58 @@
 // Names and Input
+var displayString = ""
+var backwardString = ""
+
+var hacker1 = "Randy"
+console.log("The driver's name is " + hacker1);
+
+var hacker2 = prompt("What is your name?");
+console.log("The navigator's name is " + hacker2)
+
+if (hacker1.length > hacker2.length){
+    console.log("the driver has the longest name, it has " + hacker1.length + " characters")
+}else if(hacker2.length > hacker1.length ){
+    console.log("the navigator has the longest name, it has " + hacker2.length +" characters")
+}else{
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,58 @@
...
+}else if(hacker2.length > hacker1.length ){
+    console.log("the navigator has the longest name, it has " + hacker2.length +" characters")
+}else{
+    console.log("Wow, you both got equally long name, " + hacker1.length + " characters!!")
+}
+
+for (x = 0; x < hacker1.length; x++){
+    displayString = displayString + hacker1[x].toUpperCase() 
+    + " ";
+
```
sandrabosk: This could be done by: displayString = hacker1[x].toUpperCase() + " ";
although much cleaner solution would be using these JS methods:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .join(" ")
      .toUpperCase()
  );
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,58 @@
...
+}
+console.log(displayString);
+
+for (x = hacker2.length -1 ;x >= 0;x--){
+    backwardString = backwardString + hacker2[x] ;
+
```
sandrabosk: The same as above, this works, but a better solution is using JS methods:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .reverse(" ")
      .join("")
  );
}
```

sandrabosk:
Great job Rob. Thank you very much for submitting.

PR #101, opened 2018-04-02
FTWD Miami: Ryan Martin & Diego Leal

ryancraigmartin:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
 // Names and Input
-
+var hacker1 = "Ryan";
+var hacker2 = "Diego";
```
ta-web-mia: in order to ask user for an input your would have to use prompt   var hacker2 = prompt("what is your name?")

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+else if(hacker1.length > hacker2.length){
+  console.log("The Driver has the longest name, it has " + hacker1.length + " characters.");
+}
+else if(hacker1.length < hacker2.length){
+  console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters.");
+}
+
+for(var i = 0; i < hacker1.length; i++){
+  console.log(hacker1[i].toUpperCase() + " ");
+}
```
MarcosPalaciosIII: the for loop is printing out each letter on a separate line. In order to get it to print out in one line you would try putting the for loop within a function and have a variable that would return or get logged at the end of the function.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+
+for(var i = (hacker2.length - 1); i > -1; i--){
+  console.log(hacker2[i]);
+}
```
MarcosPalaciosIII: Same as the above for loop. You would have used a function with a variable that gets called at the end of the function.

PR #103, opened 2018-04-02
MIA-Brian/Rawan - half way there

B-riJ:


ta-web-mia:
overall good job on the exercise. Remember when using code from any source that you should understand the code you are using before using it. Good job commenting where you are using code from.

PR #105, opened 2018-04-02
ftwd miami summer emison

summeremison:


ta-web-mia:
Requesting feedback from user you can use prompt.  hacker.length shouldn't be set to anything, to find out the length of a variable you have to put variableName.length so something like var driver = hacker1.length.  For printing something out to be viewed you would have to console.log() what you want in the function or console.log() the function itself.

PR #106, opened 2018-04-02
FTWD MIA Rafael Guasp lab-javascript-basic-algorithms

rguasp:


ianizaguirre:
Hi, here is some Feedback.

1) Here is some feedback for the code block below. There is no purpose for `var sameLength = false;` it does not do anything in the code, so I don't know why it was made?
```js
var hacker1 = "Alexander";
var hacker2 = "Bill";
var sameLength = false;

if(hacker1.length > hacker2.length){
  console.log("The Driver has the longest name, it has " + hacker1.length + " characters");
}
else if(hacker2.length > hacker1.length){
  console.log("Yo, navigator got the longest name, it has "+ hacker2.length +" characters");
  
} 
else {
  sameLength = true;
  console.log("wow, you both got equally long names, "+ hacker1.length +" characters!!");
}
```
* You could have just written the above code like this and got the same result:
```js
var hacker1 = "Alexander";
var hacker2 = "Bill";

if(hacker1.length > hacker2.length){
  console.log("The Driver has the longest name, it has " + hacker1.length + " characters");
}
else if(hacker2.length > hacker1.length){
  console.log("Yo, navigator got the longest name, it has "+ hacker2.length +" characters");
  
} 
else {
  console.log("wow, you both got equally long names, "+ hacker1.length +" characters!!");
}
```

2. The rest of the code:
```js
var shortest;
var same = true;

if(hacker1.length <= hacker2.length){
  shortest = hacker1.length;
}
else{
  shortest = hacker2.length;
}

for(var i = 0; i < shortest ; i++){
  if (hacker1[i].toUpperCase() < hacker2[i].toUpperCase()){
    same = false;
    break;
  }
  else if(hacker2[i].toUpperCase() < hacker1[i].toUpperCase()){
    same = false;
    break;
  }
}

if (same === true){
  console.log("What?! You both got the same name");
}
```
* The rest of that above code should have been written inside of a function. Having If/Else and Loops outside of a function or code block should be avoided since we want to avoid placing things in the Global Scope.

PR #108, opened 2018-04-02
Luis Flores

chiko15:
I worked with Rob and Randy


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,49 @@
-// Names and Input
+var hacker2 = prompt("What is your name?");
+console.log("The navigator's name is " + hacker2)
 
+if (hacker1.length > hacker2.length){
```
sandrabosk: At this point, hacker1 is not defined. So we have to define it first.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,49 @@
...
+}else if(hacker2.length > hacker1.length ){
+    console.log("the navigator has the longest name, it has " + hacker2.length +" characters")
+}else{
+    console.log("Wow, you both got equally long name, " + hacker1.length + " characters!!")
+}
 
-//Conditionals
+for (x = 0; x < hacker1.length; x++){
+    displayString = displayString + hacker1[x].toUpperCase() 
+    + " ";
```
sandrabosk: This works: displayString = hacker1[x].toUpperCase() + " ";
Although, better solution would be something like this:
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .join(" ")
      .toUpperCase()
  );
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,49 @@
...
 
+}
+console.log(displayString);
 
-// Lorem ipsum generator
+for (x = hacker2.length -1 ;x >= 0;x--){
+    backwardString = backwardString + hacker2[x] ;
```
sandrabosk: The same as up, this could work: backwardString = hacker2[x] ;
```
for (var i = 0; i < 1; i++) {
  console.log(
    hacker1
      .split("")
      .reverse(" ")
      .join("")
  );
```

sandrabosk:
Good job overall! Thank you for submitting.

PR #109, opened 2018-04-02
MIA - Claudia Summer 

clausan121:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,40 @@
...
+if (hacker1< hacker2) {
+  console.log("The driver's name is longer")
+}
+else if (hacker2 > hacker1) {
+  console.log("The navigaor has the longest name")
+}
+else if (hacker1 === hacker2) {
+  console.log("both names are long");
+}
+
```
ta-web-mia: to find out which var is longer, you just need to compare hacker1.length to hacker2.length .  there's no need to hacker.length =..

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,40 @@
-// Names and Input
+var hacker1;
+hacker1 = "claudia";
+console.log("the driver's name is" + " " + hacker1 );
+var hacker2;
+hacker2 = "summer";
```
ta-web-mia: in order to ask the user to input something you would do  var hacker2 = prompt("What is the navigator's name?");

ta-web-mia:
Few mistakes and need to prompt when asking user for input. Over all heading in the right direction.

PR #147, opened 2018-05-28
PAR Valerie and Meri

valerielinares:
@mpligioso


ta-web-paris:
You did a really great job !
You seems to have understand everything we talked about yesterday 💯 

If you want to go further, and finish the iteration 10, you can check the split method.. 

Continue like that ! 👍

PR #148, opened 2018-05-28
PAR Jana & Thierry

janamr:


starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,65 @@
...
+var str="";
+for (var i = 0; i < hacker1.length; i++) {
+  console.log(hacker1[i].toUpperCase());
+  str += hacker1[i].toUpperCase() + " ";
+}
+console.log(str);
+
+console.log(hacker2.split('').reverse().join(''));
+
+if(hacker1.toLowerCase().split('').sort() < hacker2.split('').sort()){
```
ta-web-paris: It won't work like this, as you are sorting all the letters, per exemple in Jana, as there is two "a" it will be sorted before "arthur" (only one "a") but it shouldn't ;) 

You only have to remove the .sort() method and you code will be good !

```js
var hacker1Lower = hacker1.toLowerCase();
var hacker2Lower = hacker2.toLowerCase();

if (hacker1Lower > hacker2Lower) {
console.log("The navigator's name goes first");
} else if (hacker1Lower < hacker2Lower) {
} else {
};

```

starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,65 @@
...
+
+console.log(paragraphs.split(' ').length);
+console.log(paragraphs.indexOf('et'));
```
ta-web-paris: The indexOf method allow you to catch the actual index (the position in an array) of a specific element, in your case it will output 24, because the first occurence of "et" is at the index 24 of the array.

What you can try is to create an empty array, then iterate over "paragraphs" and add an if else statement that will push "et"  whenever the loop encounters it, and then you just have to use the length method to know the actual number of element :)

ta-web-paris:
Good job, you manage to go through almost all the iterations!

PR #149, opened 2018-05-28
PAR Jihad & Diane

Dianedel:
@eloji9

We didn't get to finish the "lorem ipsum generato" part. We'll try to do it later.

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,52 @@
-// Names and Input
+//Names and Input
+console.log("I'm ready");
 
+var hacker1 = "Diane";
+console.log("The driver's name is " + hacker1);
 
 //Conditionals
+var hacker2 = prompt("what is the name of the navigator?");
+console.log("The navigator's name is " + hacker2);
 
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,52 @@
...
+  console.log("The Driver has the longest name, it has " + hacker1.length + " characters");
+} else if (hacker2.length > hacker1.length){
+  console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters");
+} else {
+  console.log("wow, you both got equally long names, " + hacker1.length + " characters!!");
+}
+
+//Loops
+var str = "";
+for (var i = 0; i < hacker1.length; i++) {
```
ta-web-paris: You don't need this, actually you can compare two string just by using the greater > operator, like this : 

```js

if (hacker1 < hacker2) {
} else if (hacker2 < hacker1) {
  console.log("The navigator's name goes first");
} else {
}

```

You can add a toUpperCase or toLowerCase method so they will be writting the same way otherwise : 
"Diane" will be sorted before "arthur".

PR #150, opened 2018-05-28
PAR Manelle, Shanshan & Nadia

ManelleG:
@moonchie
@nadiahachemi


ta-web-paris:
Good job, like the fact that you use regular expression to "clean" the text in the last iteration.

Keep up the good work! 👍

PR #151, opened 2018-05-28
PAR Ghaith & Marie

marieglr:
@GhaithHan

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,107 @@
 // Names and Input
+var hacker1 = "Marie";
+console.log("The dirver's name is " + hacker1);
 
+var hacker2 = "Ghaith";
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,107 @@
 // Names and Input
+var hacker1 = "Marie";
+console.log("The dirver's name is " + hacker1);
 
+var hacker2 = "Ghaith";
+console.log("The navigator's name is " + hacker2);
```
ta-web-paris: You can "generate" a dynamic name by asking the user : 
```js
var hacker2 = prompt("What's your name ? ")
```

ta-web-paris:
Good job! except for some small mistake I mention you did a great job, as it's just the beginning you could have wrote less "complex" stuff for the last iteration per example but otherwise 🥇

PR #152, opened 2018-05-29
PAR Jihad & Diane

eloji9:
@dianedel

We were able to finish the whole exercise this morning.
The results seem to be accurate.

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,5 +1,5 @@
 // Names and Input
-
+console.log("Hello, world!")
 
 //Conditionals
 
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,60 @@
...
+
+function wordCount(lorem) {
+  return lorem.split(" ").length;
+}
+console.log(wordCount(lorem));
+
+console.log(lorem.indexOf(" et "));
```
ta-web-paris: The indexOf method will output the index (the position) of a specific element in an array, so it's not the good way to complete the exercise (but this method will be really helpful later!)

PR #156, opened 2018-06-11
BCN - Andrea, Julien

Julienchilhagopian:
@andreaontheblock

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,44 @@
-// Names and Input
+var hacker1 = "Julien";
+
+console.log('The driver\'s name is ' + hacker1);
+
+var hacker2 = 'Andrea'; 
+
+console.log('The driver\'s name is' + hacker2);
+
+var length1 = hacker1.length;
+var length2 = hacker2.length;
+
+if (length1 > length2) {
+  console.log('The Driver has the longest name, it has ' + length1 + ' characters');
+} 
+else if (length2 > length1) {
+  console.log('Yo, navigator got the longest name, it has ' + length2 + ' characters');
+}
+else if (length1 === length2){
+  console.log('wow, you both got equally long names, ' + length1 + ' characters');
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,44 @@
-// Names and Input
+var hacker1 = "Julien";
+
+console.log('The driver\'s name is ' + hacker1);
+
+var hacker2 = 'Andrea'; 
+
+console.log('The driver\'s name is' + hacker2);
+
+var length1 = hacker1.length;
```
tawebbcn: it's ok, but you don't need to create a variable for that. It will not modify its value if you modify the ```hacker1``` variable.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,44 @@
...
+var length2 = hacker2.length;
+
+if (length1 > length2) {
```
tawebbcn: you could have used ```if (hacker1.length > hacker2.length) {```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,44 @@
...
+} 
+else if (length2 > length1) {
+  console.log('Yo, navigator got the longest name, it has ' + length2 + ' characters');
+}
+else if (length1 === length2){
+  console.log('wow, you both got equally long names, ' + length1 + ' characters');
+}
+
+
+var letters = hacker1.split('').join(' ');
```
tawebbcn: cool

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,44 @@
...
+
+function reverse (hacker) {
+  var array = hacker.split("");
+
+  array = array.reverse();
+  array = array.join("");
+
+  console.log(array);
+  return array;
+}
```
tawebbcn: nice stuff

PR #174, opened 2018-07-30
Kasia and Jen

katros:


Noelialeon:
Great job!! Please check the point number 8, right now is not working properly (It should print a name that starts with "A" before than another that starts with "B"). Take a look at the string.localeCompare() method ;)

PR #180, opened 2018-08-06
Par Paul & Julie

Youlie03:
@PaulAbarnou37

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,50 @@
 // Names and Input
+var hacker1 = "Name";
+console.log("the driver name is " + hacker1);
 
+
+var hacker2= prompt("what is your user name ?");
+console.log("the navigator's name is " + hacker2);
 
 //Conditionals
+if (hacker1.length > hacker2.length){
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,50 @@
...
+var hacker1Upp = hacker1.toUpperCase();
+console.log(hacker1Upp);
+console.log(hacker1Upp.split('').join(' '));
+
+console.log(hacker1.split("").reverse().join(""));
+
+var lexicoSort = [hacker1, hacker2];
+lexicoSort.sort();
+console.log(lexicoSort);
+if (lexicoSort[0] = hacker1 ){
```
ta-web-paris: this if statement is not working, if I enter Marie, I still get "the driver goes first" :
that is because you are using  "=" (affectation operator) in your condition instead of "===" (equality operator). This is a very common mistake !

ta-web-paris:
Good job! Good logic and great use of string methods! be careful to the difference between = and === and don't neglect for loops, as they will be super important all along the course :)

PR #183, opened 2018-08-06
PAR Abi & Manon

MSLN8:
@abisummers

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,80 @@
 // Names and Input
-
+Abi Summers Github abisummers
+Manon Salaun Github MSLN8
+// Paste your code here
+var hacker1 = "Abi"
+console.log ("the drive's name is "+ hacker1 )
+var hacker2= "Manon"
+console.log ("the navigator's name is " + hacker2 )
 
 //Conditionals
 
+if (hacker1.length>hacker2.length) {
+    console.log ("the driver has the longest name, it has " + hacker1.length)
+}
+    else if (hacker1.length<hacker2.length) {
+        console.log ("yo navigator got the longest name, it has " + hacker2.length + " characters !!")
+    }
+    
+    else if (hacker1.length===hacker2.length) {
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,80 @@
 // Names and Input
-
+Abi Summers Github abisummers
+Manon Salaun Github MSLN8
```
ta-web-paris: Be careful to always comment out any expression that is not part of your code by adding // before it, otherwise it will throw errors when you try to execute your file. 
There are different shortcuts to comment multiple lines on mac and linux, try to find them it will be very useful :) (ctrl+K,ctrl+C works on any type of file on linux!)

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,80 @@
...
+// Paste your code here
+var hacker1 = "Abi"
+console.log ("the drive's name is "+ hacker1 )
+var hacker2= "Manon"
```
ta-web-paris: You can ask the user to enter a name using the prompt command, but you did that perfectly well later with the palindrome question!

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,80 @@
...
+ console.log("Yo, the navigator goes first definitely")
+}
+else  {
+ console.log("you both have the same name")
+}
+
+var string = prompt("Write a palindrome");
+
+
+var string = string.split(" ").join("").toLowerCase();
```
ta-web-paris: 1.You shouldn't declare a variable twice, if you want to apply a method on the string you just prompted, you can just do it like that : 
var myString = prompt("Write a palindrome");
myString = myString.toLowerCase();
console.log(myString);

2. The join method is the opposite of the split method, it cancels the effect of the first method, I am not sure what you were trying to do here ? :)

3. Also, it is not advised to call a variable name by a JavaScript type name, you'll see later that it will cause problems in the code execution, you can call it for example myString :)

PR #184, opened 2018-08-06
PAR Mohamed & Constance

SMoZam:
@Cr27910

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,46 @@
 // Names and Input
 
+var hacker1 = "Mohamed";
+
+console.log("The driver's name is: " + hacker1);
+
+var hacker2 = prompt ( "What's the navigator's name : ") ; 
+
+console.log( "The navigator's name is: " + hacker2);
+
+
 
 //Conditionals
+
+var l1 = hacker1.length ;
+var l2 = hacker2.length;
+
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,46 @@
...
+
 
 //Conditionals
 
+var l1 = hacker1.length ;
+var l2 = hacker2.length;
+
+if (l1 > l2) {console.log("The Driver has the longest name, it has " +l1 +" characters"); }
+else if (l2 > l1) {console.log("Yo, navigator got the longest name, it has " +l2+ " characters"); }
+else { console.log("wow, you both got equally long names, " +l1+ " characters!!");
```
ta-web-paris: Yes, perfect logic and syntax, I would advise to always format the code with line breaks, even in if statements. 
As you get used to read code that is always formatted this way, you'll see that it makes a difference when it comes to read it fast :)
if (l1 > l2) {
    console.log("The Driver has the longest name, it has " +l1 +" characters"); 
}
else if (l2 > l1) {
    console.log("Yo, navigator got the longest name, it has " +l2+ " characters"); 
}
else { 
    console.log("wow, you both got equally long names, " +l1+ " characters!!");
 }

ta-web-paris:
Everything we saw yesterday is perfectly understood, good job !!

PR #187, opened 2018-08-13
MIA FTWD Lauren Cahill + Stephen Allison

laurencahill:


MarcosPalaciosIII:
good work on the basic algorithms lab. For the compare part you can use a function called localeCompare().  this usually works by comparing ` a.localCompare(b)` and it returns a value of 1 0 or -1. You can have that in a function to return the a or b depending on which would come first.

PR #188, opened 2018-08-13
MIA FTWD Alex-Rojas

alexrojas:
first Work

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,35 @@
-// Names and Input
+let hacker1 = "Alex"
+console.log(`The driver's name is ${hacker1}`)
+
+let hacker2 = "Eren"
+
+console.log(`The navigator's name is ${hacker2}`)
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,35 @@
...
+  console.log(x)
+
+  var y = hacker2.split('').reverse().join('')
+  console.log(y)
+
+//  console.log(y.revers())
+
+var total = [hacker1, hacker2]
+
+console.log(total.sort())
```
MarcosPalaciosIII: this function should return only 1 value (the person who comes first). One way to have done this is to do this `console.log(total.sort()[0])`. Another thing you could have done is use the localeCompare() function.

MarcosPalaciosIII:
good work on the basic algorithms lab. Make sure to check your code with instructions to assure your getting all the parameters taken care of as requested.

Keep up the good work.

PR #189, opened 2018-08-13
MIA FTWD - Eran Usha & Alex Rojas

EUsha5:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,29 @@
...
 
+let x = hacker1.toUpperCase();
+let y = x.split('');
+console.log(y.join(' '))
 
-// Lorem ipsum generator
+console.log(hacker2.split('').reverse().join(''))
+
+let total = [hacker1, hacker2];
+console.log(total.sort())
```
MarcosPalaciosIII: this function should return only 1 value (the person who comes first).  One way to have done this is to do this `console.log(total.sort()[0])`. Another thing you could have done is use the localeCompare() function.

MarcosPalaciosIII:
good work on the basic algorithms lab. Make sure to check your code with instructions to assure your getting all the parameters taken care of as requested. 

Keep up the good work.

PR #190, opened 2018-08-13
MIA full time web dev

yonathanh:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,56 @@
...
+} else {
+  console.log(
+    `yo, navigator has the longest name, it has ${hacker2.length} characters`
+  );
+}
 
 // Lorem ipsum generator
+console.log("Driver's name");
+for (let i = 0; i < hacker1.length; i++) {
+  console.log(hacker1[i].toUpperCase());
```
sandrabosk: You should make an empty space between each letter.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,56 @@
...
+}
+
+console.log("Nav's name");
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  console.log(hacker2[i]);
```
sandrabosk: This is printing all the letters in the backward order, but also you need to put it back to the string.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,56 @@
...
+}
+
+if (hacker1 < hacker2) {
```
sandrabosk: To sort the strings alphabetically, try using ```.localeCompare()``` or some other sorting method.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,56 @@
...
+} else if (hacker1 > hacker2) {
+  console.log("yo, the navigators goes first definitely");
+} else if (hacker1 === hacker2) {
+  console.log("what?! you both got the same name?");
+}
+
+let palindrom = prompt("input a word to test for palindrom", "harry potter");
+
+function checkPalindrom(palindrom) {
+  for (let i = 0, j = palindrom.length; i < palindrom.length; i++, j--) {
```
sandrabosk: The test is failing but if you make a small update where you assign value to the ```j``` it will work:
```
for (let i = 0, j = palindrom.length-1; i < palindrom.length; i++, j--) {
	// code goes here
}
```

sandrabosk:
Good work. Keep it like this. Thank you for submitting.

PR #222, opened 2018-10-13
Paris Web Dev Part Time - Sarobidy

SaroMhs:


ta-webpt-par:
Very good. You just missed the part where you are supposed to ask for the name of the second persona. To do so, you can use the method <prompt>.
Other than that, great work ! You can also try to do the bonus if you're up for it ! 
Yeah !!!
-Jihad

PR #225, opened 2018-10-15
Paris - Régis & Laura

laurastromboni:
@jourisque

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,68 @@
 // Names and Input
 
+var hacker1 = "Laura";
+console.log("The driver's name is " + hacker1);
+
+var hacker2 = "Régis";
+console.log("The navigator's name is " + hacker2);
+
+hacker1Up = hacker1.toUpperCase();
+
+var hacker1Array1 = "";
+
+console.log(hacker1Array1);
+
+var hacker1Array2 = "";
+
+console.log(hacker1Array2);
+
+var promptEndroit = "";
+var promptEnvers = "";
+
+promptEndroit = prompt("Give me a string");
+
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,82 @@
 // Names and Input
 
+var hacker1 = "Laura";
+console.log("The driver's name is " + hacker1);
+
+var hacker2 = "Régis";
+console.log("The navigator's name is " + hacker2);
```
ta-web-paris: You can interact with the user by asking his name dynamically with the prompt command : 
```
var hacker2 = prompt("What is the navigator's name?");
console.log("The navigator's name is " + hacker2);
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,82 @@
...
+else if (hacker1.length > hacker2.length){
+  console.log("The Driver has the longest name, it has " + hacker1.length + " characters");
+}
+else {
+  console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters");
+}
+
+for (i=0; i < hacker1.length; i++){
+  hacker1Array1 += hacker1Up[i] + " " ;
+}
```
ta-web-paris: good logic on this loop and on the one underneath ! You just forgot to do a console.log of hacker1Array1 and hacker1Array2 after each loop to print the result.(Also, these are not technically arrays, they are still strings :) )

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,82 @@
...
+
+if (hacker1[0] > hacker2[0]){
+  console.log("Yo, the navigator goes first definitely");
+}
+else if (hacker1[0] < hacker2[0]){
+  console.log("The driver's name goes first");
+}
+else {
+  console.log("What?! You both got the same name?");
+}
```
ta-web-paris: Here you are only comparing the first letter of each name, what if both name start with the same letter but are different names ? (like Niccolò and Nicolas  :) ) 
you can actually compare strings alphabetical order by comparing them directly with the "<" sign : 
```
if (hacker1 > hacker2) {
} else if (hacker1 < hacker2) {
} else {
}
```
You can check the fist answer here to understand how it works : https://stackoverflow.com/questions/10198257/comparing-2-strings-alphabetically-for-sorting-purposes

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,82 @@
...
+for (i=promptEndroit.length-1;i >= 0; i--){
+  promptEnvers += promptEndroit[i];
+}
+
+if (promptEndroit === promptEnvers){
+  console.log("It is a palindrome !");
+}
+else {
+  console.log("It is not a palindrome !");
+}
```
ta-web-paris: Good, your palindrom code works well !! Try to keep it a bit more organised by putting all of the pieces for this bit of logic together. Also, you dont need to initialize your promptEndroit var with an empty string before asking an input to the user : 
```
var promptEndroit = prompt("Give me a string");
var promptEnvers = "";
for ( i = promptEndroit.length - 1; i >= 0; i--) {
  promptEnvers += promptEndroit[i];
}

if (promptEndroit === promptEnvers) {
  console.log("It is a palindrome !");
} else {
  console.log("It is not a palindrome !");
}
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,82 @@
...
+
+console.log("There are " + countWords.length + " words in this string");
+
+for (i=0;i<countWords.length;i++){
+  if (countWords[i] === "et"){
+    countEt ++; 
+  }
+}
+
+console.log("There are " + countEt + " 'et' in this string.");
```
ta-web-paris: Again the code works, good job! 
You initialised the countEt variable with an empty string up there, but countEt eventually becomes a number. You would be better signaling your intent if you would initialize it with 0 : 
```
var countWords = subSections.split(" ");
var countEt = 0;

console.log("There are " + countWords.length + " words in this string");

for (var i = 0; i < countWords.length; i++) {
  if (countWords[i] === "et") {
    countEt++;
  }
}

console.log(countEt);
```

ta-web-paris:
Good job !! Your use of loops is very good and you manage to find solutions to every problems !
Be careful to the readability of your code, this is also a important part of writing good code! It would be easier for your TA's to understand your code if you would divide each exercise is different blocks that you comment instead of declaring all of you variable on the top and performing some logic already there. You can also focus on semantic naming and thinking of variable types :)
Keep up the good work !! :+1:  :+1:

PR #230, opened 2018-10-15
paris-christian-jeannicolas

blksnk:
@tidochrys

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -5,3 +5,78 @@
 
 
 // Lorem ipsum generator
+// Names and Input
+var hacker1 = "jean nicolas"
+console.log("The driver's name is " + hacker1 + ".");
+
+var hacker2 = prompt("What is the navigator's name?").toLowerCase();
+console.log("The navigators's name is " + hacker2 + ".");
+//Conditionals
+
+var hacker1Length = hacker1.length +1;
+var hacker2Length = hacker2.length +1;
+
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,78 @@
...
+};
+console.log(str1.toUpperCase());
+
+
+
+function reverseString (str2) {
+  var str2;
+  for (var i = 0; i <= (hacker2.length - 1); i++) {
+  str2 += hacker2[i];
+  };
```
ta-web-paris: This for loop is not really needed, it is the same thing of doing :
```
var str2 = hacker2;
```

ta-web-paris:
Good job !! Conditions, loops and string methods seem well understood !
Keep up the good work ! :+1:

starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,78 @@
...
+    console.log("The driver's name goes first");
+  }
+
+  else if (names[0] === hacker2) {
+    console.log("Yo, the navigator goes first definitely");
+  };
+}
+else {
+  console.log("What?! You both got the same name?");
+};
```
ta-web-paris: In the first big if statement, to check if they have the same name, you are only comparing the first letter of each name. To be sure your code also works if there names start with the same letter but have different names, you can write your code like that : 
```
if (hacker1 === hacker2){
} else if (names[0] === hacker1) {
}  else  {
} 
```

PR #232, opened 2018-10-15
Paris - Niccolo & Helen

hwoldu:
@nicrodemus

not sure that lorem ipsum exercise is correct.

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,46 @@
 // Names and Input
 
+console.log("I'm ready!");
+var hacker1 = "Helen"
+console.log("The driver's name is " + hacker1);
+var hacker2 = prompt("What is your name, navigator?");
+console.log("navigator' s name is " + hacker2);
+
+
+if (hacker1.length > hacker2.length) {
+  console.log("The Driver has the longest name, it has " + hacker1.length + " characters ")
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,46 @@
...
+console.log(spaceDriverName);
+
+var reverseNavigatorName = "";
+for ( var l = hacker2.length-1; l >= 0; l--) {
+  reverseNavigatorName += hacker2[l];
+}
+  console.log(reverseNavigatorName)
+
+console.log(hacker2.split('').reverse().join(''));
+var Order =[hacker1, hacker2];
```
ta-web-paris: Good !! Now you could just add : 
```
if (hacker1 === hacker2){
} else if (names[0] === hacker1) {
}  else  {
} 
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,46 @@
...
+console.log(loremIpsum.split(" ").length);
+console.log(loremIpsum.split("et").length-1)
```
ta-web-paris: you can put spaces around " et " in order to have only those which are not part of a word :)

ta-web-paris:
Good job !! You understand both loops and string methods, that's good!
:+1: :+1:

PR #233, opened 2018-10-15
Paris - Corrado & Chloé

griev04:
@chloehd


ta-web-paris:
Great use of loops and string method, and I like that you are careful about naming well your variables!
Keep up the good work !

PR #246, opened 2018-10-22
MIA FTWD Daniel Sauthier

DSauthier:


MarcosPalaciosIII:
Good work on the basic algorithms lab. Main take away after working in a group of 3 is that there are multiple ways of figuring challenges out. 

Keep up the good work.

PR #248, opened 2018-10-22
MIA FTWD Travis Grubbs

T-Grubbs:


MarcosPalaciosIII:
Good attempt on the basic algorithm lab. Good job thinking ahead by putting the reverse into a function so that you can use it again later if needed. In most cases It is good to have code of that style in a function so that you don't have to continually write out the same block of code over and over. 

Keep up the good work.

PR #249, opened 2018-10-22
MIA FTWD Rustam Samoigullin

Rustam1993:


MarcosPalaciosIII:
Great job on the basic algorithms lab. The main take away is that there is usually more than one way in coding to resolve a challenge. Some just take longer than others.

Keep up the great work.

PR #252, opened 2018-10-22
Assignment day 1 Dea AMS

Deakristina:


IronhackAMS:
Really nicely done. The Array.sort() method is much more eloquent! And good use of regular expression. Keep up the good work!

PR #254, opened 2018-10-22
MIA FTWD Yoli 

YoliYoli:


MarcosPalaciosIII:
Good job on the basic algorithms lab. The main take away from the lessons is that there are usually more than one way in coding to resolve a challenges. For this one when you used the for loop to reverse everything, you could have also done ``` blah.split('').reverse().join('')```  this is the same process without having to write out a for loop. 

Keep up the good work.

PR #257, opened 2018-10-22
MIA FTWD Rachel Hofer

Rachel-Hofer:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,67 @@
...
+console.log(userString);
+
+let blah3 = ''
+let blah4 = ''
+
+for(let i = 0; i < userString.length; i++) {
+  blah3 = blah3 + (userString[i])
+}
+for(let i = userString.length-1; i > -1; i--) {
+  blah4 = blah4 + (userString[i])
```
MarcosPalaciosIII: another method for doing this is ``` blah4.split(' ').reverse().join(' ')```

MarcosPalaciosIII:
Great job on the basic algorithm lab. There are some other methods you could have used other than doing for loops for majority of the challenges, but with time you will learn them. Good use of for loops. 

Keep up the great work.

PR #260, opened 2018-10-23
BCN - Selim Lebbady

lebbady:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,90 @@
...
+}
+else if (driver.length < navigator.length){
+  for (var i = 0; i <= (driver.length-1); i++){
+    if(dict[driver[i].toLowerCase()] < dict[navigator[i].toLowerCase()]){
+      i = (driver.length -1);
+      console.log("The driver's name goes first");
+    }
+    else {
+      console.log("Yo, the navigator goes first definitely");
+    }
```
tawebbcn: Amazing but there's an easier way. JS already knows the order so you can simplify your code only comparing the first two letters of the names and it will knows which goes first. Like that :
`
var hacker1Order = hacker1[0];
var hacker2Order = hacker2[0];

if (hacker1Order < hacker2Order) {
    console.log('The driver\'s name goes first');
}  etc....
`

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,90 @@
...
+  "s":19,
+  "t":20,
+  "u":21,
+  "v":22,
+  "w":23,
+  "x":24,
+  "y":25,
+  "z":26
+};
+debugger;
```
tawebbcn: Try to clean your code once you've finished. But cool, you used the debugger!!

PR #261, opened 2018-10-23
BCN - Elias Miralles

eliasMC89:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,45 @@
 // Names and Input
+var hacker1 = "Messi";
 
+console.log("The driver's name is " + hacker1);
 
-//Conditionals
+var hacker2;
 
+hacker2 = prompt("What is the navigator's name?");
```
tawebbcn: It is not mandatory to declare before the variable and then assign to that something, you can do the same thin in one line:
``` 
var hacker2 = prompt("........")
```

PR #262, opened 2018-10-23
BCN - Sergi Llonch

sllonch:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,102 @@
...
+}
+
+console.log(hacker1Capital.toUpperCase());
+
+/*
+for (var letter in hacker1) {
+        hacker1Capital += hacker1[letter];
+        hacker1Capital += ' ';
+}
+*/
```
tawebbcn: Try to clean your code, if you don't use code, maybe you can delete it

PR #263, opened 2018-10-23
BCN - Ceci Barudi

ceciliabarudi:
what i managed to finish, names-input-conditionals-loop-firstpart

Still missing: Loops and Bonus

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,22 @@
 // Names and Input
+var hacker1 = 'Cecilia';
+var hacker2 = window.prompt('What\'s the navigator\'s name?');
 
+console.log('The driver\'s name is ' + hacker1);
+console.log('The navigator\'s name is ' + hacker2);
 
-//Conditionals
+// Conditionals
+if (hacker1.length > hacker2.length ) {
+    console.log('The driver has the longest name, it has ' + hacker1.length + ' characters');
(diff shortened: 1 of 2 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,22 @@
...
 
-//Conditionals
+// Conditionals
+if (hacker1.length > hacker2.length ) {
+    console.log('The driver has the longest name, it has ' + hacker1.length + ' characters');
+} else if (hacker1.length < hacker2.length ) {
+    console.log('Yo, navigator got the longest name, it has ' + hacker2.length + ' characters');
+} else if (hacker1.length === hacker2.length) {
+    console.log('Wow, you both got equally long names, ' + hacker2.length + ' characters!!');
+};
```
tawebbcn: You can put else and not else if in this case as there are no other possibilities

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,22 @@
...
 
+// Loops
+for (var i = 0; i < hacker1.length; i++) {
+    console.log(hacker1[i].toUpperCase());
+}
```
tawebbcn: In this case you are doing a loop to display every single letter inside of the string individually, try with the join() to print all the letters together. It should be something like :
```
var addSpace = ' ';
console.log(hacker1.split('').join(addSpace));
```

PR #264, opened 2018-10-23
BCN - Jack Watkins

jackwatk:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,61 @@
+'use strict';
```
tawebbcn: Niceeee!!!

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,61 @@
...
+
 // Names and Input
+var hacker1 = "dave";
+console.log("the drivers name is " + hacker1);
+
+let hacker2 = "hello";
```
tawebbcn: Two things: 
- try to keep your code consisten, you use everywhere var and here let...
- it isn't mandatory to declare the variable before and then change it. You can do : 
```
var hacker2 = prompt('.....')
console.log('....' + hacker2)
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,61 @@
...
+    debugger;
+    //lex order
+    var alphabet=["a", "b", "c", "d", "e", "f", "g",
+    "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t",
+    "u", "v", "w", "x", "y", "z"];
+
+var driverNameStart = hacker1[0];
+var navigatorNameStart = hacker2[0];
+var alpha1 = alphabet.indexOf(driverNameStart);
+var alpha2 = alphabet.indexOf(navigatorNameStart);
```
tawebbcn: you are not using this var, maybe you can get rid of those.

PR #265, opened 2018-10-23
BCN - Kamila Purymska

KamilaPurymska:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,23 @@
...
+} else if (hacker1.length < hacker2.length) {
+    console.log('Yo, navigator got the longest name, it has ' + hacker2.length +' characters');
+} else (hacker1.length === hacker2.length); {
+    console.log("wow, you both got equally long names, " + hacker1.length + " characters!!");
+};
 
 
-// Lorem ipsum generator
+for (var i=0; i<hacker1.length; i++);
+console.log(hacker1[i].toUpperCase());
```
tawebbcn: You are not using in the right way the for loop, you are declaring it but not working inside of it: it should be 
```
for (var i=0; i<hacker1.length; i++) {
 console.log(hacker1[i].toUpperCase());
};
```

In this way you are displaying all the letters of the hacker1 name but individually, you were asked to print all the letters together divided with a space and it should be something like:  
```
var addSpace = ' ';
console.log(hacker1.split('').join(addSpace));
```

But there are many ways to do that 😄

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,23 @@
...
+
+
+
+for (var i=0; i<hacker1.length; i--);
```
tawebbcn: Same as before, you are not using the for loop.

PR #266, opened 2018-10-23
BCN - Gabriel Oliver

Gabriel0liver:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,39 @@
...
+
+console.log(hacker2Reverse);
+
+if(hacker1 > hacker2){
+    console.log("The driver's name goes first");
+}else if(hacker2 > hacker1){
+    console.log("Yo, the navigator goes first definitely")
+}else{
+ console.log("What?! You both got the same name?");
+}
```
tawebbcn: You are not checking if they are in an alphabetic order. To do that you should check the first letters of each name and compare them. It should be something like : 
```
var hacker1Order = hacker1[0];
var hacker2Order = hacker2[0];

if (hacker1Order < hacker2Order) {
    console.log('The driver\'s name goes first');
}
```

PR #269, opened 2018-10-23
BCN - Laura Enriquez

morgenmuffel93:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,53 @@
 // Names and Input
+var hacker1 = "Tania"
+console.log("The driver's name is " + hacker1);
+var hacker2 = "Anaiso"
+console.log("The navigator's name is " + hacker2);
```
tawebbcn: The purpose was to use  prompt but it's ok also like that

PR #270, opened 2018-10-23
BCN - Giulia 

MarloMorgan:


starter-code/basic-algorithms.js
```diff
@@ -1,4 +1,58 @@
...
+
+for ( let name =0; name < hacker.length; name ++){
+    let printName = hacker.toUpperCase();
+    console.log(printName);
+};
+
+for (let i = hacker.length-1; i>= 0; i--){
+    let reverseName = hacker[i];
+    console.log(reverseName);
+}
```
tawebbcn: Try with that to display the reverse name all together
 ```
var reverseName = '';
for (var ix = hacker2.length -1; ix >= 0; ix--) {
  reverseName += hacker2[ix]  
}

console.log(reverseName);
```

PR #271, opened 2018-10-23
added javascript exercises

jqbaena:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,62 @@
...
+
+if(hacker1.length > hacker2.length)
+{
+  console.log('The Driver has the longest name, it has XX characters');
+
+}else if(hacker1.length < hacker2.length){
+ console.log('Yo, navigator got the longest name, it has XX characters');
+}else{
+ console.log('wow, you both got equally long names, XX characters!!');
+
```
tawebbcn: The only thing that you are missing is to display how much they are long. You should replace the XX in the console log with the value you want to print.

PR #280, opened 2018-11-01
AMS-Luis Wilhelmi Corona

lwilhelmi:


IronhackAMS:
Nice work! Cool use of the unshift function!

PR #291, opened 2019-01-07
Paris - Christina & Céline 

frietpoune:
For the last part, we could only remove the first space and first coma. How to replace all spaces and comas? 

@CIBailey

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,65 @@
 // Names and Input
 
+console.log("I'm Ready!");
+
+var hacker1 = "Celine";
+console.log("The driver's name is " + hacker1);
+
+var hacker2 = "Christina";
+console.log("The navigator's name is " + hacker2);
 
 //Conditionals
 
+if (hacker1.length < hacker2.length){
+  console.log("Yo, navigator got the longest name , it has " + hacker2.length + " characters");
+}
+else if (hacker1.length > hacker2.length){
+  console.log("the Driver has the longest name, it has " + hacker1.length + " characters");
+} else {
+  console.log("wow, you both got equally long names," + hacker1.length + " characters");  
+}
+
+var toUpperCaseString = "";
+for(var i = 0; i <= hacker1.length-1; i++){
+toUpperCaseString += hacker1[i] + " " ;
+}
+console.log(toUpperCaseString.toUpperCase());
+
+var reverseString = "";
+for(var i = hacker2.length-1; i >= 0; i--){
+reverseString += hacker2[i];
+}
+console.log(reverseString);
+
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,65 @@
...
+} else if (sortedName[0] === hacker2){
+  console.log("Yo, the navigator goes first definitely")
+}
 
 // Lorem ipsum generator
+
+
+var userString = prompt("Enter anything");
+
+var userString = userString.replace(' ', '');
```
ta-web-paris: If you wanted to replace ALL the commas and spaces the easiest way would be with a regex (they can get SUPER confusing). 

```
var userString = userString.replace(/ /g, "");
var userString = userString.replace(/,/g, "");
```

the two / / means that what you are seeing is a regex. inside you put the code. In the first line, you can see you are replacing the space with nothing, and in the second example you can see the comma that again you are replacing with nothing. the g at the end stands for global (I think), which searches the entire string and replaces everything. 

If you wanted to, you could have done this in a for loop and checked each character to see if was a space or comma.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,65 @@
...
+var userString = userString.toLowerCase();
+console.log(userString);
+
+var reverseStringPal = "";
+for(var i = userString.length-1; i >= 0; i--){
+reverseStringPal += userString[i];
+}
+console.log(reverseStringPal);
+
+if (userString.toLowerCase === reverseStringPal.toLowerCase){
```
ta-web-paris: be careful here- you need to add () at the end of the toLowerCase 

```
reverseStringPal.toLowerCase()
```

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,65 @@
...
+}
+console.log(toUpperCaseString.toUpperCase());
+
+var reverseString = "";
+for(var i = hacker2.length-1; i >= 0; i--){
+reverseString += hacker2[i];
+}
+console.log(reverseString);
+
+var names = [hacker2, hacker1];
```
ta-web-paris: your sort function doesn't work as you think it does- 

if I put my name in as 'abi' your names appear first. but if I put 'Abi' my name is first. 

This is because it's sorting by the [code](https://www.w3schools.com/charsets/ref_html_ascii.asp) of the letter. 

you can check out this [link](https://stackoverflow.com/questions/6712034/sort-array-by-firstname-alphabetically-in-javascript) on another way to sort by name. I can take a look with you in person if you'd like!

ta-web-paris:
Good job! You seems to understand what we covered in class 🎉

PR #292, opened 2019-01-07
Paris - Christophe BEGUE & Armand BIGLARY

abiglary:
No particular comment.

@grosbegue


ta-web-paris:
What you've done is good! I like how you looked into the .reverse() method and were able to use it. You seem to have understood everything from the lesson! 

your console.log()`s are all really good- you clearly know what the console.log() is referring to from the code 🙌

PR #293, opened 2019-01-07
Paris - Thomas & Delphine

Deie:
We didn't have time to finish the bonus part of the exercice

@tomdreamer

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,51 @@
 // Names and Input
+console.log("I'm ready");
 
+var hacker1 = "Delphine";
+var hacker2 = prompt("What's the navigator's name ?");
+
+
+
+console.log("The driver's name is " + hacker1);
+console.log("The navigator's name is " + hacker2);
+
+if (hacker1.length > hacker2.length) {
+    console.log("The Driver has the longest name, it has " + hacker1.length + " characters");
+}
+else if (hacker2.length > hacker1.length) {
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,51 @@
...
+    sep += hacker1[i] + " "
+}
+console.log(sep.toUpperCase());
+
+var rev = "";
+for (var i = hacker2.length - 1; i > 0; i--) {
+    rev += hacker2[i];
+    console.log(hacker2[i]);
+}
+console.log(rev);
```
ta-web-paris: if you check your console.log() here, you'll notice that it's missing the first letter. 

You need to change your for loop slightly. You want i to be greater than OR equal to 0  `i >= 0 `

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,51 @@
...
+
+var lex = [hacker1, hacker2];
+lex.sort();
+console.log(lex.sort());
 
+if (lex.indexOf(hacker1) < lex.indexOf(hacker2)) {
```
ta-web-paris: your sort doesn't work as you expect. If you add Abi, the order is correct but if i add abi the order is wrong. This is because of the [character codes](https://www.w3schools.com/charsets/ref_html_ascii.asp)

To solve this- you can make everything the same case using .toLowerCase() or .toUpperCase()

PR #294, opened 2019-01-07
Paris - Jules & Jules

julesbadin:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,84 @@
-// Names and Input
+//Names and Input
+
+var hacker1 = "Jules";
+var hacker2 = promt("What's your name ?");
```
ta-web-paris: you mis-spelt prompt here

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,84 @@
...
+
+
+var words = latin.split(' ');
+
```
ta-web-paris: You didn't need to plus one here. But I really like how your way of counting the number of words

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,84 @@
...
+}
+  console.log(nameSpace);
+
+for (var i = hacker1.length - 1; i >= 0; i--) {
+  nameReverse += hacker1[i];
+}
+  console.log(nameReverse);
+
+
+if (hacker1[0] < hacker2[0]) {
```
ta-web-paris: for this to work properly, you need to make it case insensitive. this is because you're really comparing the [character codes](https://www.w3schools.com/charsets/ref_html_ascii.asp) which is different for uppercase and lowercase letters

ta-web-paris:
Good job!! What you've done is good, and it's cool how you used a regex - they can get super complicated 🙌

PR #296, opened 2019-01-07
Paris (dev107) Nadjie & Samuel

petitkriket:
@NadjieVal
last loop ( part 8) is unfinished, we didnt had the time to finish it properly and add the conditionnal for "Yo, the navigator goes first definitely"

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,76 +1,155 @@
-// Names and Input
-console.log("I'm Ready!");
+// 1. Create a variable `hacker1` with the driver's name
 
 var hacker1 = "Samuel"; // driver
+
+// 2. Print `"The driver's name is XXXX"` 
+console.log("The driver's name is " + hacker1);
+
+// 3. Create a variable `hacker2` and [ask the user
 var hacker2 = prompt("What's your name ?"); // navigator
 
+// 4. Print `"The navigator's name is YYYY"`
+console.log("The navigator's name is " + hacker2);
 
-// print the navigator name
-console.log(hacker2);
+//##
+/*
+5. Depending on which name is longer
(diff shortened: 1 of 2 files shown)


petitkriket:
tried a bit further on my own.. not sure it's possible for you to access the previous push did when during pair programming ?

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,155 @@
...
+//## Loops
+/*
+6. Print all the characters of the driver's name, separated by a space and [in capitals]
+ "J O H N"`
+  */
+
+var i = 0;
+var result = "";
+
+while(i < hacker1.length ) {
```
ta-web-paris: While this works, It's better to use a for loop because you're less likely to forget things. 

An alternative way to make everything uppercase, is to just do it once at the end on the final result. `result.toUpperCase()`

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,155 @@
...
+	resultTwo.push(strTwo);
+
+	int++;
+}
+
+// reverse the array order and join all the items togethers without adding a space to print jessica acissej
+var resultTwoReversed = resultTwo.reverse().join("");
+
+// uppercase last caracter regex https://stackoverflow.com/questions/32149836/how-to-capitalize-the-last-letter-of-each-word-in-a-string
+var hackerNewName = resultTwoReversed.toLowerCase().replace(/[a-z]\b/g, c => c.toUpperCase());
```
ta-web-paris: Nice!! using a regex can be super powerful but also really hard to understand so be careful when using them in your own code

ta-web-paris:
We can see both versions- so continuing on your own is not a problem :) 

What you've done is good! You have tested for different inputs like if something is lowercase which is a good skill (and important!!) to have. 

🙌

PR #298, opened 2019-01-07
Paris Raja and Aurelie

RaproCode:
@AurelieBeno


ta-web-paris:
What you've done is good! I like how you looked up something and found out about .reverse() as a way of reversing the string 👏

PR #299, opened 2019-01-07
Paris - Alexandre - Bastien

alexjoly13:
@basttran

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,3 +1,45 @@
+var hacker1 = prompt("Enter the driver's name"); console.log("The driver's name is " + hacker1); 
+
+
+var hacker2 = prompt("Enter the navigator's name"); console.log("The navigator's name is " + hacker2); 
+  if (hacker1.length === hacker2.length) {
+    console.log("wow, you both got equally long names, " + hacker1.length + " characters!!") } 
+  else if (hacker1.length <= hacker2.length) { 
+    console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters"); }
+  else {     
+    console.log("The Driver has the longest name, it has " + hacker1.length + " characters"); }
+
+
+var spaceDriver = "";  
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,3 +1,45 @@
...
+} else if (hacker1 < hacker2) {
+  console.log("The driver's name goes first");
+} else {
+  console.log("Yo, the navigator goes first definitely");
+};
+
+var alphabet = "abcdefghijklmnopqrstuvwxyz";
+var candidate = prompt("Testez la palindromie du mot");
+for (i=0 ; i <= candidate.length-1; i++) {
+  if (!alphabet.includes(candidate[i])) {candidate=candidate.substr(0,i-1)+candidate.substr(i+1,candidate.length-1);
```
ta-web-paris: I'm not entirely sure what you are trying to do here. There is an easier way to check to see if it's a palindrome

```

if (nameNormal.replace(" " ,"") === nameReversed.replace(" ", "")) {
  console.log("It's a palindrome !!");
} else {
  console.log("It's not a palindrome...");
};
```

starter-code/basic-algorithms.js
```diff
@@ -1,3 +1,45 @@
+var hacker1 = prompt("Enter the driver's name"); console.log("The driver's name is " + hacker1); 
+
+
+var hacker2 = prompt("Enter the navigator's name"); console.log("The navigator's name is " + hacker2); 
+  if (hacker1.length === hacker2.length) {
+    console.log("wow, you both got equally long names, " + hacker1.length + " characters!!") } 
+  else if (hacker1.length <= hacker2.length) {
```
ta-web-paris: here you don't want to be checking if it's greater than or equal to- you only want to check to see if it is greater than since you're checking for the longest name

ta-web-paris:
What you've done is good! You seem to understand how loops work which will be important throughout the course 💪

PR #300, opened 2019-01-07
Paris - Estelle & KevinM 

KevGhub:
Sorry, no time for Bonus

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,27 @@
 // Names and Input
 
 
+var hacker1 = "Kevin";
+console.log("The driver’s name is " + hacker1);
+var hacker2 = "Estelle";
+console.log("The navigator’s name is " + hacker2);
+
 //Conditionals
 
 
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,27 @@
...
+    console.log("The Driver has the longest name, it has " + hacker1.length + "characters");
+   }
+   else if (hacker2.length > hacker1.length ){
+   console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters");
+   }
+   else {
+   console.log("wow, you both got equally long names, " + hacker2.length + "characters!!");
+   };
+   
+   var UpperCase = ""
```
ta-web-paris: you didn't quiet finish this bit- you're not making it uppercase (yet)

you could either make it uppercase on each iteration, or do it after the loop on the final string

ta-web-paris:
What you have done is good! If you want to continue, let us know and we can give you some more feedback once you've done it :)

PR #301, opened 2019-01-07
Paris - Kevin & Estelle 2019

Wnski:
"hard git"

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,32 @@
 // Names and Input
-
+var hacker1 = "Kevin";
+console.log("The driver's name is " + hacker1);
+var hacker2 = "Estelle";
+console.log("The navigator's name is " + hacker2);
 
 //Conditionals
-
-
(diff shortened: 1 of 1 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,32 @@
...
+var UpperCase = "";
+for (i=0 ; i < hacker1.length ; i++){
+  UpperCase += hacker1[i] + " ";
+}
+console.log(UpperCase.toUpperCase());
 
 
-// Lorem ipsum generator
+var ReverseCase = "";
+for (i=0 ; i < hacker2.length ; i++){
```
ta-web-paris: your reverse doesn't quiet work. For this to work you need to loop backwards. Rather than starting on 0, you want to start on the length of the string and go down until it reaches 0. 

```
for (i= hacker2.length -1 ; i >= 0  ; i--){
  ReverseCase += hacker2[i] + " ";
} 
```

ta-web-paris:
What you've done is good! Yous seems to understand loops which will be important for later lessons 🙌

PR #350, opened 2019-03-20
PAR - Marmier Franck

1testsample:


starter-code/basic-algorithms.js
```diff
@@ -8,29 +8,84 @@ console.log(`The navigator's name is ${hacker2}`);
...
+    .join("")
+);
+
+if (hacker1.toLowerCase() < hacker2.toLowerCase())
+  console.log(`The driver's name goes first`);
+else if (hacker1.toLowerCase() > hacker2.toLowerCase())
+  console.log(`Yo, the navigator goes first definitely`);
+else if (hacker1.toLowerCase() == hacker2.toLowerCase())
+  console.log(`What?! You both got the same name?`);
+
```
ta-web-paris: This is good but the problem is that if you put everything to lowercase, you change the lexicographic value of the string. A Capital letter doesn't have the same value as a lower case letter in the lexicographic environment.

ta-web-paris:
Bravo, tu geres !

PR #355, opened 2019-03-20
PAR - Susan Almassi

sioucre:
here's my finished lab
thanks

### DIFF
--- a/starter-code/basic-algorithms.js
+++ b/starter-code/basic-algorithms.js
@@ -1,7 +1,108 @@
 // Names and Input
-
+//1.
+var hacker1 = "Juan";
+//2.
+console.log("The driver's name is " + hacker1);
+//3.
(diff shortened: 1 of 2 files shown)


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,108 @@
...
+console.log(result1);
+
+//8.
+if (hacker1[0] < hacker2[0]) {
+  console.log("The driver's name goes first");
+} else if (hacker1[0] > hacker2[0]) {
+  console.log("Yo, the navigator goes first definitely");
+} else {
+  console.log("What?! You both got the same name?");
+}
```
ta-web-paris: Ok you by passed it a bit. here you are ask to compare the lexicographic order of the two strings. Lexicographic order is a bit wired but you can check the method: localCompare !

ta-web-paris:
Great use of Regex ! Bravo

PR #356, opened 2019-03-20
PARIS javascript-basic-algorithms Miguel Bento

MiguelIronHack:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,128 @@
...
+
+/* 8. Check lexical behavior */
+if (hacker1[0].toLowerCase() < hacker2[0].toLowerCase()) {
+  console.log("The driver's name goes first");
+} else if (hacker2[0].toLowerCase() < hacker1[0].toLowerCase()) {
+  console.log('Yo, the navigator goes first definitely');
+} else {
+  console.log('What?! You both got the same name?');
+}
+
```
ta-web-paris: toLowerCase() is not relevant when it comes to lexicographic order because it change the value of the letter. Cap letter and lower case as different value. Check localCompare()

ta-web-paris:
Wp ! its great ! methods or for loops, its your choice !

PR #379, opened 2019-04-01
MiaFTWDAPR- Justin Velez

jvelez523:
Worked with Jon


ta-web-mia:
Well done! I hope you guys learned about some JavaScript methods that you didn't know about before. If there's anything worth committing to memory, it's definitely the array methods. We'll learn about and practice a little more with arrays tomorrow, so you'll definitely have a chance to get more comfortable with them.

If you guys get around to the last bonus question, keep in mind that it will be very similar to how you approached the word count one. Of course, you won't be splitting the string with an empty space. Give it some thought if you have time, and keep up the good work!

- Nick

PR #381, opened 2019-04-01
MIA FTWD APR Yulia Angel

salapjaka:
done


ta-web-mia:
Well done! One small thing that I noticed was that in some places of your code, you're declaring your variable on a line without a value, and then placing a value on it in the line immediately after:

image

There's nothing inherently wrong with doing this, but you can simplify things and just assign a value to the variable on the same line you declare it. In other words, you can just erase line 5.

Nitpicks aside, everything you have is functional and well written. If you have extra time and get the chance to work on the Lorem Ipsum bonus questions, you'll definitely have a chance to practice one of the array methods that you used earlier in the assignment. The general idea is that you would convert your string into an array by splitting the string at specific characters. Give it some thought if you get around to it. Keep up the good work!

- Nick

PR #382, opened 2019-04-01
MIA FTWD APR- Daniela Olano / Nathaniel Carvajal

dolano13:


ta-web-mia:
Good job! I did notice one issue though, and it very well may have been a placeholder while you guys were debugging. Right now, your conditional statements that compare the lengths of the two names are hardcoded for 4 characters right now:

image

The following conditional all assume 4 characters in length as well. In order to have the console.log dynamically determine the length of the name, you would want to use string concatenation to reference the length of the name and combine it with the string. Which is fancy talk for doing this:

image

Also I think you may have gotten the driver and navigator mixed up when you were referencing them in these conditionals. Definitely double check these when you get the chance and see if you can fix the issues. If you need help with any of this, definitely let me know and we can take a look at it together.

Other than that, everything else you have seems pretty spot on. If you get a chance to move onto the Lorem Ipsum bonus questions, I would suggest you use the .slice() method to determine the locations of empty spaces and "et"s in the strings. Keep up the good work!

- Nick

PR #383, opened 2019-04-02
MIA FTWB APR- Katy Brown/Noor

katybrown91:


ta-web-mia:
Well done! I don't have too much to critique here, and I really like how you declared sections of the activity as separate functions. Very compact, and very efficient!

The main takeaways from this activity were definitely the array methods, so do what you can to commit them to memory. If you guys get around to the Lorem Ipsum questions, you'll have to implement the .split() method to check for either empty spaces or "et"s in the generated strings. We'll learn more about arrays tomorrow, so take what you can into tomorrow's lesson and be prepared to learn. This course will never stop building on previous material, so it's a good mindset to get into early on. Keep up the good work!

- Nick

PR #384, opened 2019-04-02
MIA FTWD - Nour Cherif Essoussi - April 2019

Nourcs:


ta-web-mia:
Nice job! You did a good job covering weird edge cases too, like palindromes that contain spaces or different capitalizations of letters. One tiny tiny issue I found was that the word count was a little bit off. Right now you're checking for an empty space to separate the string with. However, words that are separated by a line break (new paragraphs in other words) are not going to be separated by a space. In this case, your program is counting 259 words when there are actually 261. If you can find a way to solve this minor issue that would be great, but I wouldn't fixate on it too hard.

You're definitely starting in a good place, I'm super impressed with what you've done. Keep up the good work!

- Nick

PR #385, opened 2019-04-02
MIA FTWD Jon 

jonriquer:


ta-web-mia:
Well done! There are 2 tiny edge cases that you can try and tackle if you're feeling up to it, but I wouldn't worry too much about it!

1. Right now, your palindrome checker is failing for instance where the string has capital letters and spaces. You could fix this by making all of the characters the same capitalization, and by removing the spaces from the string. 

2. Your word count is very close, but it's 1 word short of the actual count right now (338 vs. 339). The reason this is happening is because words separated with a line break that start a new paragraph aren't separated by the empty space you're checking for. If you were to remove the line breaks from the string, it would work just fine. Or if you could find some way to check for line breaks, that would be even better.

If you have time, it might be good practice to try and do these things. But since the pace won't let up, I'd just consider them for now and think about them the next time you need to do something similar. Overall, you've done a great job with this assignment. Keep up the good work!

- Nick

PR #386, opened 2019-04-02
MIA FTWD Anghy Pomar js

apomarn:


ta-web-mia:
Nice job! I'm very impressed that you covered the edge cases for the palindromes by checking for spaces and capitalization. The only small thing I see that can be fixed is that your word count is a little bit off. Right now, it's counting 347 instead of 349. The reason for that is that while it's checking for words separated by spaces, it isn't checking for words separated by line breaking. Since there are 2 line breaks separating the paragraphs, 2 words are going uncounted. I wouldn't fixate on it too hard, but consider approaching this if you have extra time.

Other than that, everything is flawless. Keep up the good work!

- Nick

PR #387, opened 2019-04-02
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Well done! You did a good job checking for capitalization and spaces in the palindrome question, so kudos to you for being thorough. If you have extra time and try to approach the Lorem Ipsum questions, I would suggest you make use of the .split() method to check for certain characters in the string. Everything is totally functional and clean, so I have no complaints. Keep up the good work!

- Nick

PR #397, opened 2019-05-28
Paris FTWD - Camille&Camille - Basic algorithms done

camcharp:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+    console.log("wow, you both got equally long names, " + hacker1.length + " characters!!");
+} else {
+    console.log("Yo, navigator got the longest name, it has " + hacker2.length + " characters");
+}
 
+// Print name
+let lettres = ""
+
+for(i = 0; i < hacker1.length; i++) {
+    console.log(lettres += hacker1.charAt(i).toUpperCase() + " ")
```
ta-web-paris: Care, here you are doing a console log inside a loop, so you will print something at every round of the loop. It is a good idea to start from letter which is empty and add letters to it ! The console log should be right after ;)

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+}
+
+for(i = hacker2.length; i >= 0; i--) {
+    console.log(lettres += hacker2.charAt(i))
```
ta-web-paris: Here, lettres has kept in memory the hacker1 letters because the scope is the same and you use +=. Maybe try this in to separate functions or assign a new variable, or start back letter empty :-)

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,63 @@
...
+
+let counter = 1;
+for (i=0; i < text.length; i++) {
+ if (text.charAt(i) === " ") {
+   counter += 1;
+ }
+}
+console.log("The text contains " + counter + " words.");
+
+var n = text.search("et");
```
ta-web-paris: POUR .search() / Si la recherche aboutit, search() renvoie un entier qui correspond à l'indice de la première correspondance trouvée dans la chaîne. Du coup pas sur que ca soit l'outil adequate ici ;)
Une facon de le faire
` var count = 0;
 for (let i = 0; i < loremIps.split(" ").length; i++) {
   if(loremArr[i] === "et") {
     count += 1
   }
 }`

ta-web-paris:
Bon travail dans l'ensemble ! Well done !

PR #413, opened 2019-06-10
MIA-FTWD Michael Roth with Mitch Malinin

mikerawth:


ta-web-mia:
Nice job, everything is correct! Just a soft reminder, but the word count function on the 3 paragraphs of text has to check for more than just a space to count a word. In addition to an empty space, line breaks (\n) are also used to separate words between paragraphs. An edge case like this is nothing to worry too much about though. Keep up the good work!

Nick

PR #417, opened 2019-06-10
MIA-FTWD Sergui Morejon with Sergio Velikopoljski

sergui20:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,66 @@
...
+
+Nulla eleifend elit vitae pretium facilisis. Vestibulum accumsan sed leo at molestie. Nullam feugiat convallis porttitor. Vivamus pulvinar lectus vitae maximus dapibus. Vivamus et orci justo. Etiam sit amet ullamcorper velit, vitae congue urna. Morbi ut nulla ipsum. Mauris elit arcu, congue at venenatis quis, tincidunt quis sapien. Nam nunc est, aliquet in interdum in, rhoncus in eros. Donec lobortis ipsum ac dignissim aliquet. Sed ac lorem elementum, commodo purus eget, finibus ipsum.`;
+
+let splitParagraph = paragraph.split(' ')
+console.log(splitParagraph.length)
+
+let etCounter = 0;
+
+splitParagraph.forEach( word => {
+    if (word === 'et') {
```
MarcosPalaciosIII: Good job but to get a more precise number you can use .includes("et")

MarcosPalaciosIII:
Great work on the basic algorithm lab. The main takeaway is that when working with a partner or a group of people, you may find that others can have a way of doing things that you would not think about. In turn you get another point of view in how to solve a problem.

Keep up the great work.

PR #419, opened 2019-06-10
MIA-FTWD-<Artem with Nicolle>

arstrel:
Prompt commented out


MarcosPalaciosIII:
Excellent work on the basic algorithm lab. The main takeaway is that when working with a partner or a group of people, you may find that others can have a way of doing things that you would not think about. In turn you get another point of view in how to solve a problem.

Keep up the excellent work.

arstrel:
I appreciate the immediate and positive feedback 👍 kudos go to Nicolle, my
dynamic navigator lighting the way. Great session today

PR #421, opened 2019-06-10
MIA-FTWD-Ashley-with-Sams

Ashleydarrue:


MarcosPalaciosIII:
Good work on the basic algorithm lab. The main takeaway is that when working with a partner or a group of people, you may find that others can have a way of doing things that you would not think about. In turn you get another point of view in how to solve a problem.

Keep up the good work.

PR #422, opened 2019-06-10
MIA _FTWD Matt Larrubia with William Bennett

mlarrubia:


starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,72 @@
...
+
+    console.log("wow, you both got equally long names," +" "+ hacker1.length +" "+ "characters!!");
+    
+}
+      else if(hacker2.length > hacker1.length){
+        console.log("Yo, navigator got the longest name, it has" +" "+ hacker2.length +" "+ "characters");
+    } 
+    
+    else if(hacker1.length > hacker2.length) {
+        console.log("The Driver has the longest name, it has"+  hacker1.length +"characters");
```
ta-web-mia: Not a  huge deal, but don't forget to add spacing between the strings and values that you're concatenating (to improve readability). For example, if hacker1 was the longest name, it would print:

 "The Driver has the longest name, it has4characters"

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,72 @@
...
+
+if (hacker1.localeCompare(hacker2)==0){
+    console.log("You Guys got the same name");
+}
+
+if(hacker1.localeCompare(hacker2)==-1){
+    console.log(hacker1+"'s name goes first");
+}
+
+if (hacker1.localeCompare(hacker2)==1){
```
ta-web-mia: Nice use of localeCompare! The alternative is to loop through the strings and compare the characters directly, and this is much more efficient.

starter-code/basic-algorithms.js
```diff
@@ -1,7 +1,72 @@
...
+    console.log("Yo, the navigator goes first definitely ")
+}
+
+
+let thisIsAString = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus pretium tempor blandit. Donec eu pretium ligula. Integer faucibus libero magna, nec rhoncus nunc aliquam a. Aliquam in placerat magna, et convallis risus. Ut id nibh nisl. Pellentesque consequat nunc luctus, vulputate quam eu, fermentum massa. Suspendisse ut eros ultrices, elementum tortor id, tempus libero";
```
ta-web-mia: Just a reminder, in case you were having trouble with linebreaks. You can use backticks `like this` to allow line breaks in the middle of strings, so that you can add multiple paragraphs.

ta-web-mia:
Nice work! Take a look at my feedback and let me know if you have any questions. Keep up the good work 👍 
- Nick

PR #423, opened 2019-06-10
MIA-FTWD-Nicolle/Artem

nicollegonzalez:


MarcosPalaciosIII:
Excellent work on the basic algorithm lab. The main takeaway is that when working with a partner or a group of people, you may find that others can have a way of doing things that you would not think about. In turn you get another point of view in how to solve a problem.

Keep up the excellent work.

PR #425, opened 2019-06-10
MIA-FTWD Gulfiya Isyangulova with Billy 

GulfiyaIsyangulova:


starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,43 @@
...
+}
+​
+let spaceName = "";
+for (i=0; i<hacker1.length; i++) {
+ spaceName = spaceName + hacker1[i] + " ";
+ console.log(spaceName.toUpperCase())
+}
+​
+let reverseName = hacker2.split("").reverse()
+ console.log(reverseName.join(""));
```
ta-web-mia: Correct! But just want to make the point that the .join method can be put on line 33 after the .reverse. It might look a little weird on a single line, but it does mean you can console.log JUST the variable.

starter-code/basic-algorithms.js
```diff
@@ -5,3 +5,43 @@
...
+​
+let lexiOrder = hacker1.localeCompare(hacker2)
```
ta-web-mia: Nice use of localCompare. Very perceptive to know that you didn't have to loop.

ta-web-mia:
Nice work! If you get a chance to work on the bonus iteration, keep in mind that the palindrome will borrow some of what you did on the string reverse section. Keep it up 👍

PR #449, opened 2019-08-07
PAR-FTWD-Niraja

niraja426:


ta-web-paris:
good job.
On the word count, you are off by 1, by counting with spaces you don't take into account the last word :) 
So you should initialize the count to 1.

PR #451, opened 2019-08-07
PAR FTWD Sara 

SebHacker:


ta-web-paris:
in the bonus iteration you are not executing the toUpperCase function : ).
Rest is working : )

PR #458, opened 2019-08-11
PAR FTWD - Basic Algorithms - Aimee and Margaux

aimeestone:


ta-web-paris:
you forgot to make a prompt for the hacker2 name,
it's undefined : )
result = window.prompt(message, default);
https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt

var spaceCapitals = hacker1
  .toUpperCase()
  .split("")
  .join(" ");
var upperCaseSeparatedHacker1 = "";

for (var i = 0; i < hacker1.length; i++) {
  console.log(spaceCapitals);
}

you don't need a for loop, your string has already been transformed.

//Reverse name
var reverseOrder = "";
for (var i = hacker2.length; i >= 0; i--) {
  console.log(hacker2[i]);
}

you reversed the name, but you're logging the string one at a time and not creating a new reversed string.

PR #501, opened 2019-10-21
MIA FTWD Grant Delhaes

gdel21:


ta-web-mia:
Nice work! Hopefully you have a better grasp of string and array methods that weren't necessarily a focus in the pre-work. Keep up the good work 👍 

- Nick

PR #506, opened 2019-10-21
MIA FTWD - October - Nicolas Bared

nbared:


ta-web-mia:
Nice work! If you're interested in the last bonus problem, it can be solved using the substr method or using regex match. Not super necessary, but wanted to let you know in case you wanna take a look later. Keep it up 👍 
- Nick

PR #509, opened 2019-10-21
FTWD Johnny Nguyen

jnguyen3436:


October212019.js
```diff
@@ -0,0 +1,46 @@
...
+spaceName +=(capitalName[i] + " ")
+
+}
+return spaceName
+
+}
+
+function backwardsName(name){
+let backward=""
+for(i=0;i<=name.length-1;i++){
```
ta-web-mia: I like this solution! The array.reverse() method is a great way of going about this as well.

ta-web-mia:
Nice job! Keep up the good work 👍 
- Nick

PR #513, opened 2019-10-21
MIA FTWF Vitalii Agapov

g1yk:


index.js
```diff
@@ -0,0 +1,107 @@
...
+
+wordsCounter(generatedText);
+
+// et Count:
+
+const etCounter = (text) => {
+  let result = 0;
+
+  for (let i = 0; i < text.length; i++) {
+    if (text[i - 1] == ' ' && text[i] == 'e' && text[i + 1] == 't' && text[i + 2] == ' ') {
```
ta-web-mia: Very clever! To make things easier on yourself, you can use the substr() method to do essentially the same thing. More info can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

ta-web-mia:
Nice work! If you have any questions, just let me know. Keep up the good work 👍 
- Nick

PR #515, opened 2019-10-21
FTWD MIA - October Avrahm Kleinholz

avrahmk:
Day 1 Lab 👍


index.js
```diff
@@ -0,0 +1,54 @@
...
+
+let word = "et";
```
ta-web-mia: Make sure you check for the uppercase "Et" as well, since JavaScript differentiates between uppercase and lowercase letters.

PR #524, opened 2019-10-25
[PTMIA102019] Ashraf and Kevin

alforsii:
Lab - Team work branch.


MarcosPalaciosIII:
Great job on the basic algorithms lab. You will find that there are many ways to finding the solution for problems when coding. As you get more experience you will be able to do this in other ways that you may have known about now.

Keep up the great work.

alforsii:
👍

PR #525, opened 2019-10-25
[PTMIA102019] Ashraf (or Ashrafzhon Kurbonaliev)

alforsii:
Self branch - my branch.


MarcosPalaciosIII:
Great job on the basic algorithms lab. You will find that there are many ways to finding the solution for problems when coding. As you get more experience you will be able to do this in other ways that you may have known about now.

Also looking at each commit I noticed that you had everything done. But looking at your all your commits together only shows about 25 lines of code. You seem to have deleted some stuff when you should have left it all on the file

Keep up the great work.

alforsii:
got you.

PR #526, opened 2019-10-26
[PTMIA102019] CarlosOsorio&HumbertoBuniotto

karosu:


MarcosPalaciosIII:
Good attempt on the basic algorithms lab. You will find that there are many ways to finding the solution for problems when coding. As you get more experience you will be able to do this in other ways that you may have known about now.

Keep up the good work.

karosu:
Thank you for your feedback!

I have finished all the iterations for this lab,  should I make another pull request? @MarcosPalaciosIII

PR #573, opened 2019-11-18
ING-Susanne-Malte

MSHSusanneV:


js/index.js
```diff
@@ -1,7 +1,48 @@
...
 
 
 // Iteration 2: Conditionals
+if (hacker1.length > hacker2.length){
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`);
+}
+else if (hacker1.length < hacker2.length){
+    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters.`);
+}
+else {
```
MarcoSantonastasi: prefer explicit condition like " === "

js/index.js
```diff
@@ -1,7 +1,48 @@
...
+  
+    upperSpaceHacker1 = upperSpaceHacker1 + upperHacker1.charAt(i) + " ";
+  }
+console.log (upperSpaceHacker1);
+// console.log(upperSpaceHacker1.length);
 
-// Iteration 3: Loops
\ No newline at end of file
+let reverseHacker2 = "";
+for (let i=hacker2.length; i>=0; i--){
```
MarcoSantonastasi: ERROR: loop index is outside of range correct to i = hacker2.length - 1

PR #575, opened 2019-11-18
ING-Alina-Jürgen-Sergej

JuergenGooss:
First Iteration done

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,15 +9,24 @@ console.log(`The driver´s name is ${hacker1}`);
 
 //1.3 Create a variable hacker2 with the navigator's name.//
 
-let hacker2 = "Alina";
-let hacker3 = "Sergej";
+let hacker2 = "AlinaSergej";
 
 //1.4 Print "The navigator's name is YYYY".//
 
-console.log(`The navigators are ${hacker2}, ${hacker3}`);
+console.log(`The navigator´s name is ${hacker2}`);
 
 /*Iteration 2: Conditionals
 2.1. Depending on which name is longer, print: - The Driver has the longest name, it has XX characters or - Yo, navigator got the longest name, it has XX characters or - Wow, you both got equally long names, XX characters!*/
+
+if (hacker1.length >=hacker2.length){
+  if (hacker1.length == hacker2.length){
+    console.log(`Wow, you both got equally long names, ${hacker1.length} characters!`)
+  } else {
+    console.log(`The Driver has the longest name, it has ${hacker1.length} characters`)
+  } 
+} else {
+  console.log(`Yo, navigator got the longest name, it has ${hacker2.length} characters`)
+}
 
 /*Iteration 3: Loops
 3.1 Print all the characters of the driver's name, separated by a space and in capitals i.e. "J O H N"*/
--- a/index.js
+++ b/index.js
@@ -10,7 +10,80 @@ console.log(`The driver´s name is ${hacker1}`);
 //1.3 Create a variable hacker2 with the navigator's name.//
 
 let hacker2 = "AlinaSergej";
+/*Iteration 1: Names and Input
+1.1 Create a variable hacker1 with the driver's name.*/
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -0,0 +1,116 @@
...
+let hacker2 = "AlinaSergej";
+
+//1.4 Print "The navigator's name is YYYY".//
+
+console.log(`The navigator´s name is ${hacker2}`);
+
+/*Iteration 2: Conditionals
+2.1. Depending on which name is longer, print: - The Driver has the longest name, it has XX characters or - Yo, navigator got the longest name, it has XX characters or - Wow, you both got equally long names, XX characters!*/
+
+if (hacker1.length >=hacker2.length){
```
MarcoSantonastasi: could be greatly simplified by extracting the condition as an expression to use in the if else branching

index.js
```diff
@@ -0,0 +1,116 @@
...
+  if (hacker1.length == hacker2.length){
+    console.log(`Wow, you both got equally long names, ${hacker1.length} characters!`)
+  } else {
+    console.log(`The Driver has the longest name, it has ${hacker1.length} characters`)
+  } 
+} else {
```
MarcoSantonastasi: prefer an explicit condition. see above comment

index.js
```diff
@@ -0,0 +1,116 @@
...
+  console.log(`Yo, navigator got the longest name, it has ${hacker2.length} characters`)
+}
+
+/*Iteration 3: Loops
+3.1 Print all the characters of the driver's name, separated by a space and in capitals i.e. "J O H N"*/
+
```
MarcoSantonastasi: prefer use of loops as per lesson instruction

index.js
```diff
@@ -0,0 +1,116 @@
...
+let uppercasedDriver=hacker1.toUpperCase();
+/*let split="";
+. Solution:
+for (let i = 0; i<hacker1.length; i++){
+   split += `${uppercasedDriver[i]}  `;
+  
+}
+console.log(split);*/
+
+let splittedName = uppercasedDriver.split("");
```
MarcoSantonastasi: for native string methods prefer method concatenation

index.js
```diff
@@ -0,0 +1,116 @@
...
+let nameFinal = splittedName.join(" ")
+console.log(nameFinal)
+
+//3.2 Print all the characters of the navigator's name, in reverse order. i.e. "nhoJ"//
+
+
+let reversedName = hacker2.split("").reverse().join("")
```
MarcoSantonastasi: prefer loops as per lesson instructions

index.js
```diff
@@ -0,0 +1,116 @@
...
+console.log(reversedName)
+
+//3.3 Depending on the lexicographic order of the strings, print: - The driver's name goes first. - Yo, the navigator goes first definitely. - What?! You both got the same name?//
+
+let abc = "abcdefghijklmnopqrstuvwxyz";
```
MarcoSantonastasi: ERROR: this is not the correct solution because it does not respect the lexicographic order

index.js
```diff
@@ -0,0 +1,116 @@
...
+
+console.log(`The navigator´s name is ${hacker2}`);
+
+/*Iteration 2: Conditionals
+2.1. Depending on which name is longer, print: - The Driver has the longest name, it has XX characters or - Yo, navigator got the longest name, it has XX characters or - Wow, you both got equally long names, XX characters!*/
+
+if (hacker1.length >=hacker2.length){
+  if (hacker1.length == hacker2.length){
+    console.log(`Wow, you both got equally long names, ${hacker1.length} characters!`)
+  } else {
```
MarcoSantonastasi: prefer explicit condition

index.js
```diff
@@ -0,0 +1,116 @@
...
+    console.log(`The Driver has the longest name, it has ${hacker1.length} characters`)
+  } 
+} else {
```
MarcoSantonastasi: prefer explicit condition

PR #576, opened 2019-11-18
ing-karin-dirk solution

kaalaska:


js/index.js
```diff
@@ -1,7 +1,48 @@
...
 
 // Iteration 2: Conditionals
+let lenHacker1= hacker1.length;
+let lenHacker2= hacker2.length;
 
+if(lenHacker1>lenHacker2){
+  console.log(`The driver has the longest name, it has ${lenHacker1} characters.`);
+}else if(lenHacker1<lenHacker2){
+  console.log(`It seems that the navigator has the longest name, it has ${lenHacker2} characters.`);
+}else{
```
MarcoSantonastasi: prefer explicit condition like " === "

js/index.js
```diff
@@ -1,7 +1,48 @@
...
+  console.log(`Wow, you both have equally long names, ${lenHacker2} characters!`);
+}
 
-// Iteration 3: Loops
\ No newline at end of file
+
+// Iteration 3: Loops
+let upperHacker1=hacker1[0];
```
MarcoSantonastasi: ERROR: Wrong assumption that hacker1[0] is already uppercase

js/index.js
```diff
@@ -1,7 +1,48 @@
...
+for(let i=1;i<lenHacker1;i++){
```
MarcoSantonastasi: ERROR: Loop should start at i = 0. See above

js/index.js
```diff
@@ -1,7 +1,48 @@
...
+  reverseHacker2+=hacker2[lenHacker2-i-1];
+}
+console.log(`"${reverseHacker2}"`);
+*/
+
+if(hacker1<hacker2){
+  console.log(`The driver's name goes first`);
+}else if(hacker1>hacker2){
+  console.log(`Yo, the navigator goes first definitely`);
+}else{
```
MarcoSantonastasi: prefer explicit condition like " === "

js/index.js
```diff
@@ -1,7 +1,48 @@
...
+
+/*
+reverseHacker2="";
+for(let i=0;i<lenHacker2;i++){
+  reverseHacker2+=hacker2[lenHacker2-i-1];
+}
+console.log(`"${reverseHacker2}"`);
+*/
+
+if(hacker1<hacker2){
```
MarcoSantonastasi: prefer localeCompare() as per example in the lesson material

PR #608, opened 2020-01-06
FT WD - Berlin Jan. 2020 - Vasiliki & Ciwan

ciwan-aram:


js/index.js
```diff
@@ -1,7 +1,49 @@
...
 // Iteration 2: Conditionals
 
+// Iteration 3: Loops
+
+console.log("I'm ready");
+let hacker1 = "Ciwan";
+console.log("The driver's name is " + hacker1);
+let hacker2 = "Vasiliki";
+console.log("The navigator's name is " + hacker2);
+
```
fonsogms: Nice! All good on first iteration :D
As well if you guys want to try other way to put variables in a string you could try the `` quotes where you could write your string like this `The driver´s name is ${hacker1}` as an example.

js/index.js
```diff
@@ -1,7 +1,49 @@
...
+    "It seems that the navigator has longest name, it has " +
+      hacker2.length +
+      " characters."
+  );
+} else {
+  console.log(
+    "Wow, you have both equally long names, " + hacker1.length + " characters!"
+  );
+}
+
```
fonsogms: All good in the hood guys!
Only comment about this iteration is the same as the comment above regarding the variables in a string :D

js/index.js
```diff
@@ -1,7 +1,49 @@
...
 
-// Iteration 3: Loops
\ No newline at end of file
+if (hacker1.localeCompare(hacker2) === -1) {
+  console.log("The driver´s name goes first.");
+} else if (hacker1.localeCompare(hacker2) > 0) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?");
+}
```
fonsogms: Nice job! No more comments

PR #610, opened 2020-01-06
[BET-FT] Jason Franklin, Nora Nicklis 

Norizinus:
Done

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,78 @@
 // Iteration 1: Names and Input
 
+let hacker1 = "Nora";
+//console.log(`The driver's name is ${hacker1}`);
+let hacker2 = "Jason";
+//console.log(`The navigators's name is ${hacker2}`);
 
 // Iteration 2: Conditionals
 
+if (hacker1.length > hacker2.length) {
+  console.log(
+    `The driver has the longest name, it has ${hacker1.length} characters`
+  );
+} else if (hacker1.length < hacker2.length) {
+  console.log(
+    `It seems that the navigator has the longest name, it has ${hacker2.length} characters`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker2.length} characters!`
+  );
+}
+
+let upperCaseName = "";
+for (let i = 0; i < hacker1.length; i++) {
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,78 @@
...
+  );
+} else if (hacker1.length < hacker2.length) {
+  console.log(
+    `It seems that the navigator has the longest name, it has ${hacker2.length} characters`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker2.length} characters!`
+  );
+}
```

js/index.js
```diff
@@ -1,7 +1,78 @@
 // Iteration 1: Names and Input
 
+let hacker1 = "Nora";
+//console.log(`The driver's name is ${hacker1}`);
+let hacker2 = "Jason";
+//console.log(`The navigators's name is ${hacker2}`);
```

js/index.js
```diff
@@ -1,7 +1,78 @@
...
+  );
+}
 
-// Iteration 3: Loops
\ No newline at end of file
+let upperCaseName = "";
+for (let i = 0; i < hacker1.length; i++) {
+  upperCaseName += hacker1[i].toUpperCase() + " ";
+}
+console.log(upperCaseName);
```

js/index.js
```diff
@@ -1,7 +1,78 @@
...
+
+console.log(
+  hacker2
+    .split("")
+    .reverse()
+    .join("")
+);
```
DalinaLeila: 👍 Nice

js/index.js
DalinaLeila: Looking below that you already understand how .split() and .join() works. What you can do here is simply 

`hacker1.split("").toUpperCase().join(" ")`

js/index.js
```diff
@@ -1,7 +1,78 @@
...
+
+let reverseName = "";
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  reverseName += hacker2[i];
+}
+console.log(reverseName);
+
+// Iteration 3: Loops
+
+if (hacker1 < hacker2) {
```
DalinaLeila: I would be careful to do this when you want to sort by lexicographical order because it does work for every case. Best is to use localeCompare()

PR #611, opened 2020-01-06
done Ozhan and Jonathan

JonathanSaudhof:


js/index.js
```diff
@@ -1,7 +1,58 @@
 // Iteration 1: Names and Input
 
+let hacker1 = 'Jonathan';
+console.log("Driver's name is:" + hacker1);
+
+let hacker2 = 'Ozhan';
+console.log('Navigators name is:' + hacker2);
```
fonsogms: nice job guys!
First comment:
I would recommend to keep consistency in the quoting for the strings which means, if you use this (" ") quotes use it for all the strings, or the same with the other type of quotes.
Also I would reccomend to try the 
```
`${variable}`
``` 
type of quoting. So in this case it would be 
```
`Drivers name is: ${hacker1}`
```
as an example

js/index.js
```diff
@@ -1,7 +1,58 @@
...
+} else if (hacker1.length < hacker2.length) {
+  console.log(
+    'The navigator has the longest name, it has' +
+      hacker1.length +
+      'characters.',
+  );
+} else {
+  console.log('Wow, you both have equally long names, XX characters! ');
+}
+
```
fonsogms: All good here

js/index.js
```diff
@@ -1,7 +1,58 @@
...
 
-// Iteration 3: Loops
\ No newline at end of file
+if (test1 === test2) {
+  console.log('What?! You both have the same name?');
+} else if (test1 < test2) {
+  console.log("The driver's name goes first.");
+} else {
+  console.log('Yo, the navigator goes first definitely');
+}
```
fonsogms: Nice job guys but in this case for the lexicographic order I recommend you using this:
```
if (hacker1.localeCompare(hacker2) === -1) {
  console.log("The driver´s name goes first.");
} else if (hacker1.localeCompare(hacker2) > 0) {
  console.log("Yo, the navigator goes first definitely.");
} else {
}
```
I would recommend you to check out  how does localeCompare(): 
https://www.w3schools.com/jsref/jsref_localecompare.asp

PR #612, opened 2020-01-06
[BER-FT-012020]) - Jacqueline C. & Shahinya H.

chenjacqueline:
LAB pair programming, w1d1

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,81 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
 
+// Iteration 1: Names and Input
+let hacker1 = "Jackie";
+console.log(`"The driver's name is ${hacker1}"`);
+
+let hacker2 = "Shahinya";
+console.log(`"The navigator's name is ${hacker2}"`);
 
 // Iteration 2: Conditionals
-
+if (hacker1.length > hacker2.length) {
+  console.log(
+    `The driver has the longest name, it has ${hacker1.length} characters.`
+  );
+} else if (hacker2.length > hacker1.length) {
+  console.log(
+    `It seems that the navigator has the longest name, it has ${hacker2.length} characters.`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker1.length} characters!`
+  );
+}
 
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,81 @@
...
+// 3.2 Print all the characters of the navigator's name, in reverse order.
+let reversedName = "";
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  reversedName += hacker2[i];
+}
+console.log(reversedName);
+
+// 3.3 Alphabetical order
+let hackerNames = [hacker1, hacker2];
+hackerNames.sort();
```
DalinaLeila: I would be careful to use sort when wanting to sort by lexicographical order. This will not always work because it's not actually sorting depending on the alphabet. Best is to use localeCompare().

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+if (hackerNames.indexOf(hacker1) < hackerNames.indexOf(hacker2)) {
+  console.log("The driver's name goes first.");
+} else if (hackerNames.indexOf(hacker2) < hackerNames.indexOf(hacker1)) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?");
+}
+
+// Bonus 1 Word count
+let paragraphs = "";
```
DalinaLeila: you are missing the paragraph text here :)

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+
+let numberOfWordsArr = paragraphs.split(" ");
```
DalinaLeila: At the moment you are splitting an empty string

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+  secondHalf += phraseWithoutSpaces[i];
+}
 
-// Iteration 3: Loops
\ No newline at end of file
+if (firstHalf === secondHalf) {
+  console.log("This string is a palindrome!");
+} else {
+  console.log("This string is not a palindrome.");
+}
```
DalinaLeila: Looks good! Nice that you went for the bonus 💯!

PR #617, opened 2020-01-06
[BER-FT-012020]) - Saeka I.

saekaisozumi:
LAB 01/06 d1w1

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,4 +1,52 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
+
+let hacker1 = "Tim"
+console.log("The driver's name is " + hacker1);
+
+let hacker2 = "Saeka"
+ console.log("The navigator's name is " + hacker2);
+
+if  (hacker1.length > hacker2.length) {
+console.log("The driver has the longest name, it has " + hacker1.length + " characters")
+} else if (hacker1.length < hacker2.length) {
+  console.log("It seems that the navigator has the longest name, it has " + hacker2.length + " characters")
+} else {
+  console.log("wow, you both have equally long names, " + hacker2.length + " characters!")
+}
+
+let driverUpperCase = hacker1.toUpperCase();
+console.log(driverUpperCase[0] + " "+ driverUpperCase[1]+ " "+driverUpperCase[2]);
+
+console.log(hacker2[4]+hacker2[3]+hacker2[2]+hacker2[1]+hacker2[0]);
+
+
+lexValue = hacker1.localeCompare(hacker2);
+
+if (lexValue == 0) {
+  console.log("What?! You both have the same name?");
+} else if (lexValue == 1) {
+  console.log("The driver's name goes first.");
+} else {
+  console.log("Yo, the navigator goes first definitely.")
+}
+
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,55 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
+
+let hacker1 = "Tim"
+console.log("The driver's name is " + hacker1);
+
+let hacker2 = "Saeka"
+ console.log("The navigator's name is " + hacker2);
```
DalinaLeila: This is good! What is usually done is to write this using the dollar signs and backticks. This way you do not need to use the plus sign and your code becomes cleaner. 
e.g.
``console.log(`The driver's name is ${hacker1}`)``

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+
+if  (hacker1.length > hacker2.length) {
+console.log("The driver has the longest name, it has " + hacker1.length + " characters")
+} else if (hacker1.length < hacker2.length) {
+  console.log("It seems that the navigator has the longest name, it has " + hacker2.length + " characters")
+} else {
+  console.log("wow, you both have equally long names, " + hacker2.length + " characters!")
```
DalinaLeila: See above.

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+}
+
+let driverUpperCase = hacker1.toUpperCase();
+console.log(driverUpperCase[0] + " "+ driverUpperCase[1]+ " "+driverUpperCase[2]);
```
DalinaLeila: Writing it this way will not allow your code to be flexible. This means that if your hacker1 changes his name to more than three letters, the other letters will be upperCase but will not have a space between them. 

What you can do is to loop over the "driverUpperCase" string and then add a space in every iteration. Another way (and this is much simpler) is to use the .split() and .join() method. 

In this case, you would write:

`driverUpperCase.split("").join(" ")` --> meaning you want to **split** the string at every character. Then this creates an array that you want to **join** with space.

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+
+console.log(hacker2[4]+hacker2[3]+hacker2[2]+hacker2[1]+hacker2[0]);
```
DalinaLeila: See above.

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+
+
+lexValue = hacker1.localeCompare(hacker2);
```
DalinaLeila: Nice! Make sure to declare the variable lexValue using var,let or const

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+
+But I must explain to you how all this mistaken idea of denouncing pleasure and praising pain was born and I will give you a complete account of the system, and expound the actual teachings of the great explorer of the truth, the master-builder of human happiness. No one rejects, dislikes, or avoids pleasure itself, because it is pleasure, but because those who do not know how to pursue pleasure rationally encounter consequences that are extremely painful. Nor again is there anyone who loves or pursues or desires to obtain pain of itself, because it is pain, but because occasionally circumstances occur in which toil and pain can procure him some great pleasure. To take a trivial example, which of us ever undertakes laborious physical exercise, except to obtain some advantage from it? But who has any right to find fault with a man who chooses to enjoy a pleasure that has no annoying consequences, or one who avoids a pain that produces no resultant pleasure?`
+
+console.log(`The length is ${text.length}.`);
```
DalinaLeila: This is not the number of words, but just the length of the entire text :)

js/index.js
```diff
@@ -1,7 +1,55 @@
...
+
+let counter = 0
+
+for (let i = 0; i < text.length; i ++) {
+  if (text[i] == "e" && text[i + 1] == "t"){
+    counter += 1
+  }
+}
+
```

PR #629, opened 2020-01-07
[Ber-FT]Finished Sebastian - Martin

MngSebastian:


js/index.js
```diff
@@ -1,7 +1,90 @@
...
+console.log(hacker1Capital);
+
+
+let hacker2Backwards = "";
+for (let i = hacker2.length; i > 0; i--) {
+
+    hacker2Backwards += hacker2[i - 1];
+}
+console.log(hacker2Backwards);
+
```
fonsogms: All good in the hood!
A small comment:
For the reverse loop, instead of doing hacker2[i-1] you could change it in for (let i = hacker2.length - 1; i >= 0; i--);
which in terms of common practice in js would be that way, but it still works and there is no problem with your code.
Another small note: When naming variables for loops such as "let i" remember you can use other names for the variables in the loops such as "let j".

js/index.js
```diff
@@ -1,7 +1,90 @@
...
+    }
+    if (words[i] == " ") {
+        wordCounter++;
+    }
+}
+
+console.log("number of et: " + counter);
+console.log("number of words: " + wordCounter);
+
+
```
fonsogms: Nice job guys!
I guess you guys were counting "of" instead of "et" but looks like you got the idea of the lab.
A small note on the "of" counter: If there was a word like "software" your counter will count it as an extra "of" when actually there isn´t. So a small correction would be to make sure there is a space after the "f" in your code

MngSebastian: That's actually what we went for,  we decided that we should count every "of" regardless if it is found in word or by itself.

I guess we could do if (words[i] == "o" && words[i + 1] == "f" && words[i + 1] == " ") . wright?

MngSebastian: Or should it actually be : if (words[i] == "o" && words[i + 1] == "f" && words[i + 2] == " ") .   ?

PR #632, opened 2020-01-07
[BER-FT] Karolina and Lluis

karo-cz:


js/index.js
```diff
@@ -1,7 +1,128 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
 
+//INTERATION 1
+//introduction
 
-// Iteration 2: Conditionals
+let hacker1 = "Karolina";
+let hacker2 = "Lluis";
+console.log("The drivers's name is " + hacker1);
```
itstheandre: In this situation, you could have used template literals.

Things like 
```
`The driver's name is  ${hacker1}`
```

PR #634, opened 2020-01-07
[BRL-FT-01.2020_VICTORIA_BLATKOWSKA&ELSA FILLIPIDOU] 

VictoriaBlat:
First Lab

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,71 @@
 // Iteration 1: Names and Input
+console.log("I'm ready!");
 
+let hacker1 = "Victoria";
+
+console.log(`"The driver's name is ${hacker1}."`);
+let hacker2 = "Elsa";
+
+console.log(`"The navigator's name is ${hacker2}."`);
 
 // Iteration 2: Conditionals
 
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,71 @@
 // Iteration 1: Names and Input
+console.log("I'm ready!");
 
+let hacker1 = "Victoria";
+
+console.log(`"The driver's name is ${hacker1}."`);
```
itstheandre: In this case and in the hacker2 version the double quotes are unnecessary :) the backticks are already giving you a string. :)

js/index.js
```diff
@@ -1,7 +1,71 @@
...
+
+let splitted = loremIpsum.split(" ");
+
+let arrayCounter = splitted.length;
+
+console.log(arrayCounter);
+console.log("'et' appears " + loremIpsum.match(/et/g).length + " times.");
```
itstheandre: You didn't need to use any Regex , but nicely done!

ElsaFil:
@itstheandre thanks for the review! 🤓

PR #638, opened 2020-01-08
[PAR-WDFT]-PASCAL-BESSON-FRANCK-TREBOIT ongoing

pascalbes:


TatijanaRajcic:
Good job guys. 

Some tips on some functions:

- compareNameLength
You don't really need to use the function nameLength: name1.length is shorter to write that nameLength(name1) hehe
A easiest way to write a variable inside a string would be to use, instead of "" the back ticks `` and then refer to your variable with ${variable}.
Ex: return `The driver has the longest name, it has ${nameLength(name1)} characters`;

- reverse function
Don't forget to put it in a function. Console log is not enough :)

You are on the right way for bonuses but small things are missing:

- countEt
Your function does not count the "et" that are followed by "," or ".", because, while using split with " ", it cuts those words as "et," or "et." 
Do you have an idea on how to do it? :)

- isPalindrome 
The function should work even with punctuation. "Amor, Roma" is indeed a palindrome. So the function should return true :). Tip: you should get rid of capital letters (because in a palindrome "B" = "b" which is not the case when you compare letter by letter, and the punctation gets on your way)

I am sure you can fix the bonuses and it'll be perfect! :) Good job!

PR #639, opened 2020-01-08
[PAR-WDFT-0120] Etryhard Bibby - ongoing

PierreEtryhard:


Jepassaispar:
Great job guys,
Just last iteration your function was giving the name that goes last insted of the one that goes first, but it's a detail you got everything :)
I especially like the reverseName function, very neat !

PR #642, opened 2020-01-08
[PAR-WDFT] - Anthony B. & Shannon ongoing

abrebion:


TatijanaRajcic:
Great jobs guys :)
- I like the fact that you tried different type of loops, for example while doing the first iteration in part 3
- Regarding the function compareNames: localeCompare works! Do you have an idea how you would have done it with a loop? If you are interested in it, try and do it, or come and see me and we can discuss the approach :)
- It's cool that you did the bonuses as well. If you want to go further with Bonus 1, here is a tip:
It should work with every case of "et", even when it's followed by a "." or ",". Which is not the case right now since, in those cases, the "et" will be cut by your split function as a "et." or "et,"
Nice job :)

PR #645, opened 2020-01-08
[PAR-WDFT] - Nina et Mehmet - ongoing

Gorlanova:
Finished the exercises

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,110 @@
-// Iteration 1: Names and Input
+// === TEST
+console.log("I'm ready !");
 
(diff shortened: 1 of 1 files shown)


Jepassaispar:
Great job, very cool functions, I really liked the fact that you use the same array for 2 iterations, reuse of code is always good ;)

Gorlanova:
Thank you !
For the last bonus we used “join” and I saw afterwards that we were asked not to, so we could’ve used a loop to concat all the letters of the string backwards into a new string, but we got the point ! We’ll be more careful next time

Thanks for your feedback !!

PR #646, opened 2020-01-08
[PAR-WDFT-0120]-ANDY TISBA-PAUL DICKERSON ongoing

pauldkn:


TatijanaRajcic:
Hey guys! Great job :) everything looks on point! 
- To order the namesArray, I see that you are using .sort() on both arrays, it's a good idea! There is also a build-in function called localeCompare. You could also simply compare the word themselves, since Javascript will tell you that "marie" < "paul". There is also a way to do it with loops (which is harder, but you can try if you are interested in it)
- Regarding the loremIpsum, be careful that the "et" that will end a sentence (so which will be followed by a "." or ",") will not be counted as "et" with your method. It shouldn't be hard to fix :)
- The next step for you would have been to put every code in a function. But it was not asked so it's fine :) I'm just saying since I see that you got it all!
Good job guys!

PR #651, opened 2020-01-09
[PAR-FT-012020] - Maria.MUNIZAGA & Yuval.BU ongoing

mmunizaga:


TatijanaRajcic:
Hey guys!
Your work is great!
Be careful that variables should start with small letters (LexicOrder should be lexicOrder). Another way of doing it is using the fact that in Javascript, "maria" < "youval". so you could do some conditionnal statements with this too.
I like the fact that you used functions in the last questions :) It's cleaner! 
For the bonus 1 : be careful, "et." or "et," are not counted. Regex are sometimes mysterious hehe. It would have been possible to do it even simplier. For example, by spliting by " ", and then looking for "et" "et." and "et,"
Great job guys :)

PR #652, opened 2020-01-09
Done with the first 2 iterations.

generationyct:
Done with the first 2 iterations.


generationyct:
There is going to be a new pull request which is a little more advanced.

PR #654, opened 2020-01-09
Done

fhoman:


js/index.js
```diff
@@ -1,7 +1,108 @@
...
+
+hacker1Uppercase = hacker1.toUpperCase()
+nameWithSpaces = '';
+for (let i = 0; i < hacker1Uppercase.length; i++) {
+console.log(hacker1Uppercase[i] );
+nameWithSpaces += `${hacker1Uppercase[i]} `;
+nameReverseOrder = '';
+}
+console.log(nameWithSpaces);
+for (let j = hacker2.length -1; j > 0; j-- ) {
```
LloydC: in the for loop, you should use "j>=0" otherwise the first letter won't print.

js/index.js
```diff
@@ -1,7 +1,108 @@
...
+}
+
+// Iteration 3: Loops
+
+hacker1Uppercase = hacker1.toUpperCase()
+nameWithSpaces = '';
+for (let i = 0; i < hacker1Uppercase.length; i++) {
+console.log(hacker1Uppercase[i] );
+nameWithSpaces += `${hacker1Uppercase[i]} `;
+nameReverseOrder = '';
```
LloydC: If you're not using it inside the loop, It's better to define nameReverseOrder in the global scope.

js/index.js
```diff
@@ -1,7 +1,108 @@
...
+nameReverseOrder += `${hacker2[j]}`;
+
+}
+
+console.log(nameReverseOrder);
+
+// Lexical order
+
+
+if (hacker1 === hacker2) {
```
LloydC: I think you meant to use "hacker1.length === hacker2.length"

js/index.js
```diff
@@ -1,7 +1,108 @@
...
+
+console.log(`What?! You both have the same name?`);
+
+}
+
+else if (hacker1.localeCompare(hacker2) === -1) {
```
LloydC: Again you can use the ".length" property here

js/index.js
```diff
@@ -1,7 +1,108 @@
...
+  
+console.log(`The driver's name goes first`);
+}
+  
+else if (hacker1.localeCompare(hacker2) === 1) {
+
```
LloydC: Again you can use the ".length" property here

PR #656, opened 2020-01-10
Finished Iteration 1 - 3 and Bonus 1 - By Markie Mark

generationyct:


generationyct:
Found a better way to find a substr with match()

PR #657, opened 2020-01-10
Module1_Assignment_Boyuan_Amsterdam_pt0107

sbyshjd:


js/index.js
```diff
@@ -1,7 +1,129 @@
...
+
+//B1.3 use the 'et' string and substr method to test how many 'et' word we have
+let etNum = 0;
+for (let i=0; i < outPut.length; i++) {
+  if(outPut.substr(i,4)===' et '|| outPut.substr(i,4)===' et.') {
+    etNum++;
+  }
+}
+ for(let i=0; i<outPut.length; i++) {
+   if(outPut.substr(i,3)==='Et '){
```
LloydC: You can also use another "| || operator in the first for loop and remove the second for loop.

js/index.js
```diff
@@ -1,7 +1,129 @@
...
+
+//B2.2 make a new variable to change the character order in reverse.
+let newPhrase = '';
+for(let i =phraseToCheck.length-1; i > -1; i--) {
+  newPhrase += phraseToCheck[i]
+}
+//console.log(newPhrase);
+
+//B2.3 test if the phrase is a Palindrome phrase.
+if(newPhrase === phraseToCheck) {
```
LloydC: You should use ".toLowerCase()" in your if statement in case there are uppercase letters in phraseToCheck.
So if(newPhrase.toLowerCase() === phraseToCheck.toLowerCase()){...}

PR #659, opened 2020-01-10
First pull request (1st class, lot of testing around)

matteoazzola23:
It looks like I made it. It took me a bit to upload it and master Terminal, but I believe I am getting there. I also have a couple of questions to discuss together :)

See you!

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,95 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
 
+// 1.1 Create a variable `hacker1` with the driver's name.
+
+// 1.2 Print `"The driver's name is XXXX"`.
+
+// 1.3 Create a variable `hacker2` with the navigator's name.
+
+// 1.4 Print `"The navigator's name is YYYY"`.
+
+// Iteration 1: Names and Input
+let hacker1 = "Matteo"
+let hacker2 = "Arnold"
+  console.log('The driver\'s name is ' + hacker1 + '.');
+  console.log('The driver\'s name is ' + hacker2 + '.');
+
+
+// 2.1. Depending on which name is longer, print:
+// - The driver has the longest name, it has XX characters. or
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,95 @@
...
+// 1.3 Create a variable `hacker2` with the navigator's name.
+
+// 1.4 Print `"The navigator's name is YYYY"`.
+
 // Iteration 1: Names and Input
+let hacker1 = "Matteo"
+let hacker2 = "Arnold"
+  console.log('The driver\'s name is ' + hacker1 + '.');
+  console.log('The driver\'s name is ' + hacker2 + '.');
+
```
LloydC: You can use backticks and template strings here. Like `The driver's name is ${hacker1}`

js/index.js
```diff
@@ -1,7 +1,95 @@
...
+for (i=0; i<hacker2.length; i++) {
+// let arr= [hacker2];
+  console.log(hacker2[i].toUpperCase() + "\n");
+}
+
+// Reverse name?
+// hacker1 = hacker1.split("");
+// hacker1 = hacker1.reverse();
+// hacker1 = hacker1.join("");
+
```
LloydC: You can also use a for loop which read the string starting at the end. Like for( i = hacker1.length -1; i >=0; i++){console.log(hacker[i])}

js/index.js
```diff
@@ -1,7 +1,95 @@
...
+console.log(wordCount(lorem));
+
+let num_matches = lorem.match(/et/gi).length;
+console.log(num_matches); 
+
+// bonus 2
+
+phraseToCheck ='radar'
 
-// Iteration 3: Loops
```
LloydC: One way to check if something is a palindrome is to have a variable that will store the reversed version of the string. Then you compare the original string and reversed string to see if they're the same. If yes then it's a palindrome

PR #660, opened 2020-01-10
Done

danielhoogerwerf:


js/index.js
```diff
@@ -1,7 +1,96 @@
...
+
+for (let i = 0; i < hacker1.length; i++) {
+  driverName = driverName + hacker1[i].toUpperCase() + " ";
+}
+
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  navName += hacker2[i];
+}
+
+if (hacker1.localeCompare(hacker2) === 1) {
```
lotofcaffeine: Just to improve legibility, sometimes it is good to see if it is easier using a switch. 💅🏼
`switch (hacker1.localeCompare(hacker2)) {
  case 1:
     console.log("The driver's name goes first. ")
    break;
  case 0:
     console.log("What?! You both have the same name?")
     break;
  case -1: 
    console.log("Yo, the navigator goes first definitely.")
    break;
}`

js/index.js
```diff
@@ -1,7 +1,96 @@
...
+  reversePhrasePart += firstPhrasePartClean[i];
+}
 
-// Iteration 3: Loops
\ No newline at end of file
+if (firstPhrasePartClean === reversePhrasePart) {
+  console.log("It's a palindrome!");
+} else {
+  console.log("Too bad.. no palindrome...");
+}
```
lotofcaffeine: Well done! Really great that you made even the bonus 👏🏻
Now, You just learned about functions! It would be amazing if you could practice this knowledge in this exercise, you would see how many codes you will not repeat.
Go #DRY

js/index.js
danielhoogerwerf: Aah makes sense! Thanks for the tip! 😃

PR #662, opened 2020-01-10
[AMS-PT-112020] - Rajmohan B

Rajmohankct2003:
JavaScript Basic Algorithms - Lab

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,21 +1,22 @@
-// Iteration 1: Names and Input
+console.log("Iteration 1 : ");   
 // 1.1 Create a variable `hacker1` with the driver's name.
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,118 @@
...
+
+console.log("No of words in the paragraph is : " + para1.split(" ").length);
+
+console.log("\nBonus 2 : ");
+let phraseToCheck = "Amor, Roma";
+let num2 = phraseToCheck.length - 1;
+let num1;
```
lotofcaffeine: Well done, Your code is really good. 🚀
The only point that I would consider is to give better names for your variables, just to improve legibility but it is already really awesome.

PR #663, opened 2020-01-11
[AMS]mahmut sarcan

mhmtdl:


js/index.js
```diff
@@ -1,7 +1,38 @@
...
+    else{
+      console.log(`Wow,you both have equally long names,${hacker1.length}characters!`);}
 
-// Iteration 3: Loops
\ No newline at end of file
+
+//Iteration 3: Loops   
+let name='';
+for(let i=0;i<hacker1.length;i++){
+  name +=hacker1[i].toUpperCase()
```
LloydC: You should also add an empty space: So name +=hacker1[i].toUpperCase() + " "

PR #667, opened 2020-01-13
done

valmgisbert:
I'll dO THE BONUS 2 LATER I AM T I R ED


js/index.js
```diff
@@ -1,7 +1,41 @@
-// Iteration 1: Names and Input
+//it 1 
+var hacker1 = 'Jason';
+console.log(`"The driver's name is ${hacker1}"`);
```
tawebbcn: If you use the string interpolation, you don't need the quote inside, but you solve that later in the exercises, so good :)

js/index.js
```diff
@@ -1,7 +1,41 @@
...
+  console.log(`The driver has the longest name, it has ${hacker1.length} characters.`);
+}
+else if (hacker1.length < hacker2.length) {
+  console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters.`);
+} else if (hacker1.length < hacker2.length) {
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
 
+//it 3
+for (i = 0; i <= hacker1.length - 1; i++) {
```
tawebbcn: you need to assign i to a variable, var i = 0

PR #668, opened 2020-01-13
[BCN-FTWD-012020] Stefan Berkenhoff

StefanK2ff:


js/index.js
```diff
@@ -1,7 +1,76 @@
...
+- What?! You both have the same name?*/
+
+if (hacker1 > hacker2) {
+  console.log("The driver's name goes first.")
+} else if (hacker1 < hacker2) {
+  console.log("Yo, the navigator goes first definitely.")
+} else {
+  console.log("What?! You both have the same name?")
+};
+
```
tawebbcn: The code works perfectly, but it evaluates if it's in the reverse alphabetical order. To get it in the alphabetical order, you should just need to check if (hacker1 < hacker2) in the first condition, and if (hacker1 > hacker2) in the second.

PR #670, opened 2020-01-13
FT-WEB-BCN-NEHA

annkay108:


js/index.js
```diff
@@ -1,7 +1,79 @@
...
+var loremIpsumLength = loremIpsum.length;
+var count = 1;
+for (var k= 0; k<loremIpsumLength;k++){
+  if(loremIpsum[k]==" "){
+    count+=1;
+  }
+}
+console.log(count);
+console.log((loremIpsum.match(/et/g) || []).length);
+
```
tawebbcn: Here you are checking all consecutive letters "et" that appear in the text. If you just want to search for the word "et", you can add spaces around "et":
console.log((loremIpsum.match(/ et /g) || []).length);

js/index.js
```diff
@@ -1,7 +1,79 @@
...
+var phraseToCheck = "race car";
+var phraseLength = phraseToCheck.length;
+var newPhrase = "";
+for (var l=0;l<phraseLength; l++){
+  if (phraseToCheck[l]!=" " && phraseToCheck[l]!="," &&phraseToCheck[l]!="?" && phraseToCheck[l]!="!"){
+    newPhrase+=phraseToCheck[l];
+  }
+}
```
tawebbcn: to avoid hardcoding each character you need to remove (" ", "," or "?") to get the "newPhrase", you could use regex like you did above, combining it with the string method .replace():
newPhrase = phraseToCheck.replace(/[^A-Za-z0–9]/g, "")

PR #671, opened 2020-01-13
[BCN - FTWD - 012020 - Sara del Valle]

LolaEnBeta:
Hi all!! :)

Here I let my exercise for the first day. Hope everything will be ok, and I'll be glad to know your feedback. 

Thank you so much!

Sara.


tawebbcn:
Hey Sara! Very good job. Your code is very easy to read. All your logic is very good.
The only weird thing i saw is the way you work with displaying strings. You did this:
**"Wow, you both have equally long names,", driverLength, "characters!"**
But you could have also done it like this:
**"Wow, you both have equally long names: " + driverLength + " characters"**
Or in ES6 with string interpolation:
**`Wow, you both have equally long names: ${driverLength} characters`**

Very good job!!

LolaEnBeta:
Thank you so much!! I take note of this! =)

PR #673, opened 2020-01-14
[BCN-FT-012020]-Tarek

Tking1606:
Day 1 exercise JS.


js/index.js
```diff
@@ -1,7 +1,68 @@
...
+    result1=result1+" "+hacker1[i].toUpperCase();    
+}
+console.log(result1);
+result2="";
+for (var i=hacker2.length-1;i>=0;i--){
+    result2=result2+" "+hacker2[i].toLowerCase();
+}
+console.log(result2);
+
+for(var i=0;i<=hacker1;i++){
```
tawebbcn: In your loop, you should initiate it this way (var i=0; i<hacker1.length; i++)

js/index.js
```diff
@@ -1,7 +1,68 @@
...
+        }
+        else if(hacker1[i]<hacker2[i]){
+            console.log("The driver's name goes first.");
+        }
+        else{
+            continue;
+        }
+
+    }
+}
```
tawebbcn: For this iteration, it doesn't work properly, you should compare each letter with the localCompare method

js/index.js
```diff
@@ -1,7 +1,68 @@
...
+else{
+    console.log(`Wow, you both have equally long names, ${hacker2.length} characters!.`);
+}
 
+// Iteration 3: Loops
+var result1="";
+for (var i=0;i<hacker1.length;i++){
+    
+    result1=result1+" "+hacker1[i].toUpperCase();
```
tawebbcn: To avoid to reapeat, you can use the operator += like the following example:
result1 += " "+hacker1[i].toUpperCase();

PR #699, opened 2020-01-21
[LIS JAN 20 - Antonio Penalva, Diana Odiaka]

adwpenalva:
@ironhackLis-webTeam TAs, we couldnt make the iteration 3 work.
Every time we got the result it was in different lines, we couldnt make it all in the same line.


soniameller:
Hey, we worked on the solution in class. If you still have any doubt please let us know!

PR #735, opened 2020-02-10
LIS-PTWD-AndreR

andreElCoder:
Done


ironhack-LIS-PTWD:
Thank you Andre,
good job!
Paula )

PR #739, opened 2020-02-10
LIS-PTWD-FEB

migueltomarnogueira:
@ironhack-LIS-PTWD

Work in progress. Since I'm still doing the prework I will continue this exercise later on.

- Names and Input was easy.
- I know I'm cheating on conditionals, i should be using if else statements right? :)
- I don't know how even to start the loops and the bonus, but after i do the prework till the end of the week I'll get there again.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,11 @@
 // Iteration 1: Names and Input
-
+let hacker1 = "Roadie";
+console.log("The driver's name is", hacker1);
+let hacker2 = "Passenger"
+console.log("The navigator's name is", hacker2);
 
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hey Miguel, 
Good job getting the length of the the string. It is asking you to use conditionals, and so try to implement this :) 
You have created the variables, and so you can now compare lengths using the if/else. 
E.g: 
if (hacker1.length > hacker2.length){ 
    console.log('.......')
}_ 
Definitely continue with the prework lessons, that will give you some more clarity with loops! 
Good luck :)

PR #741, opened 2020-02-10
LIS - PTWD - FEB - Julia Carvalho

juliamendesc:
@ironhack-LIS-PTWD
Did Lab up to bonus 1, could not understand how to get to bonus 2 nor how to make reverse without function. Please assist :)

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,50 @@
 // Iteration 1: Names and Input
-
+let hacker1 = "John";
+console.log(`The driver's name is ${hacker1}.`);
+let hacker2 = "Manel";
+console.log(`The navigator's name is ${hacker2}.`);
 
 // Iteration 2: Conditionals
-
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hi Julia,
well done for reaching the bonus. 
the fact that you are using a function is really good and makes a lot of sense!
if you wanted to solve this part with a for loop it could look like this: 

```
let reverse = ""
for(let i = hacker1.length -1; i >= 0; i--){
     reverse += hacker1[i]
}
console.log( reverse)

```

To give you a tip for Bonus 2:
you need to:

- create a second variable for the reversed version of the word
- find a way to remove the spaces of the both strings 
&
- bring both strings to lower or upper Case
- then compare StringToCheck with the variable you created  

let know if that helps,

we will also happily provide you with the solution code if you like but since you are on a good way I thought to give you some hints first. 

good luck!

Paula :)

PR #742, opened 2020-02-11
LIS-PTWD-FEB-MargaridaPascoal

mindfoo:
I tried not using split and join but ended up using split and join. Oh well...

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,130 @@
-// Iteration 1: Names and Input
+
+console.log("I'm ready!");
+
+/*
+Iteration 1: Names and Input
+1.1 Create a variable `hacker1` with the driver's name.
+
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hi Margarida, 

well done ! It's not a problem at all to use split() and join(). There are always several solutions to a lab and you should use the one that you understand and that works best for you. However where you are comparing the two words 
`palindromeCheck === phraseToCheck`
bear in mind the spaces and case sensitivity for both strings. Otherwise javascript won't evaluate them as the equal. 

> "Atack Cats" =/= "Atac kCats"

  but 

> "stackcats" === "stackcats"

mindfoo:
Right. Got it. Would have to evaluate each character's case as well then...
Will recheck that.
Thanks for that feedback!

PR #743, opened 2020-02-11
LIS - PTWD - FEB - Rui Pereira

RuiMPer:
@ironhack-LIS-PTWD

all iterations done, going for the bonus.

The code work, but I still don't understand 100% what this expressions mean and what they exactly say to the computer to do. in this case I think that the "+=" is confusing me.
driverName += hacker1.charAt(i)
navigatorName += hacker2[i]

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,20 @@
 // Iteration 1: Names and Input
+const hacker1 = "Rui"
+console.log(`The driver's name is ${hacker1}`)
+//-------Iteration 1----------     
+const hacker1 = "Rui";
+console.log(`The driver's name is ${hacker1}`);
 
+const hacker2 = "Miguel";
+console.log(`The navigator name is ${hacker2}`)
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hi Rui,

well done for finishing the lab!

let's look at an example regarding your question: 

```
let characters = "Abc";
console.log(characters)
// output => "Abc"
characters = "def"
console.log(characters)
// output => "def"
```

in this case you are reassigning a new value to the variable
however with += you are adding to the variable instead

```
let characters = "Abc";
characters += "def"
console.log(characters)
// output => "Abcdef"
```

so `characters += "def"` is essentially the same as 
`characters = characters + "def"` only shorter

So in relation to the lab: 
In each iteration of the for loop you are adding to the variable `navigatorName` character by character `hacker2[i]`. 

Hope this helps,
Paula :)

PR #745, opened 2020-02-12
LIS - WDPT - FEB - WILSON CRUZ

galatoide:
Done 
@ironhack-lis-ptwd


ironhack-LIS-PTWD:
Hi Wilson,
really well done! 
For the last 2. Bonus have a look at this solution: 
```//change around phraseToCheck to text results
const phraseToCheck = "Amor, Roma";

// 1st Remove spaces from String
let removedSpaces = "";
for(i= 0; i < phraseToCheck.length; i++){
  if(phraseToCheck[i] === " "){
    continue
  } else {
    removedSpaces += phraseToCheck[i].toLowerCase()
  }
}
// 2nd compare every index from front with every index from end
let output = ""
for(i = 0; i < removedSpaces.length; i++) {
  if(removedSpaces[i] === removedSpaces[removedSpaces.length-1-i]) {
    output = `${phraseToCheck} is a Palindrome. :) `
  } else {
    output = `${phraseToCheck} is not a Palindrome. :( `
    break }
}
console.log("Bonus 2: ",output)````

In iteration 3.3 the lab wants you to sort the name by alphabetical order. I would recommend you to use the method  `.localeCompare()` check out how to use it:
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare](url)

PR #775, opened 2020-03-23
 BER-FTWD-Victoria-Sami

victoria-toria:


js/index.js
```diff
@@ -1,7 +1,48 @@
...
+console.log(driverName.slice(1));
+
+let navigatorName = "";
+for (let i=hacker2.length-1; i >= 0; i--){
+  navigatorName += hacker2[i];
+}
+console.log(navigatorName);
+
+let arrayOfNames =[hacker1, hacker2];
+arrayOfNames.sort();
```
fenrew: This can give  a bug especially when it comes to capitalization of letters because a != A. The best method here would be a .localeCompare() method where a == A to prevent any bugs arising from capitalization of letters in a name.

PR #776, opened 2020-03-23
BER-FTWD-Yosra-Marie

bl00dymarie:
We (@yosrayousfi) finished the lab 1 "lab-javascript-basic-algorithms" succesfully :tada: !

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,3 +1,5 @@
+console.log("I'm ready")
+
 // Iteration 1: Names and Input
 
 
--- a/js/index.js
+++ b/js/index.js
@@ -1,9 +1,63 @@
 console.log("I'm ready")
 
 // Iteration 1: Names and Input
-
+const hacker1 = "Marie";
+console.log(`"The driver's name is ${hacker1}"`);
+const hacker2 = "Yosra";
+console.log(`"The navigator's name is ${hacker2}"`);
 
 // Iteration 2: Conditionals
-
+if(hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`)
+} else if(hacker1.length < hacker2.length) {
+    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters.`)
+} else {
+    console.log(`Wow, you both have equally long names, ${hacker1} characters!`)
+}
 
 // Iteration 3: Loops
+let newDriverName = "";
+for (let i = 0; i < hacker1.length; i++) {
+    newDriverName += hacker1[i].toUpperCase() + " "
+}
+console.log(newDriverName);
+
+let newNavigatorName = "";
+for (let i = hacker2.length-1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,63 @@
...
+}
+console.log(newDriverName);
+
+let newNavigatorName = "";
+for (let i = hacker2.length-1; i >= 0; i--) {
+    newNavigatorName += hacker2[i]
+}
+console.log(newNavigatorName);
+
+if (hacker1.localeCompare(hacker2) === -1){
```
fenrew: A few things here with locale compare:
Some browsers can return -2 and some -1 (info can be found at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeComparet) so it is more safe to write:
if (hacker1.localeCompare(hacker2) < 0) 

firstVariable.localeCompare(secondVariable) returns a negative number if the firstVariable is longer and

js/index.js
```diff
@@ -1,7 +1,63 @@
...
+    console.log(`The driver's name goes first.`)
+} else if ((hacker2.localeCompare(hacker1) === 1)){
```
fenrew: hacker1.localeCompare(hacker2) written above returns -1 if hacker1 goes first
hacker2.localeCompare(hacker1)  written here returns 1 if hacker1 goes first.

Since you changed your localeCompare AND what it returns you basically have the same condition on both of these conditions, meaning hacker2 will never go first:)

js/index.js
```diff
@@ -1,7 +1,63 @@
...
+// Bonus 1
+const paragraph = "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
+console.log(paragraph.split(" ").length);
+console.log(paragraph.split(" et ").length-1);
 
+// Bonus 2
+const phraseToCheck = "A man, a plan, a canal, Panama!";
+let phraseOnlyLetters = "";
+for (let i = 0; i < phraseToCheck.length; i++){
+    if(phraseToCheck[i] == " " || phraseToCheck[i] == "," || phraseToCheck[i] == "!" || phraseToCheck[i] == "?" || phraseToCheck == "." || phraseToCheck == "'" || phraseToCheck == '"' || phraseToCheck == "," ){
```
fenrew: So technically it can be done like this, keep however in mind that there are A LOT of special characters out there. Another way of solving this can then be to say something like:
const allLetters = abcdefghijklmnopqrstuvwz
if(allLetters.indexOf(phraseToCheck[i].toLowerCase()) > -1)

This is just an idea, your solution still works for these strings given in the labs

PR #777, opened 2020-03-23
BER-FTWD-Clara-Phillip

phpaul89:
done

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,78 @@
 // Iteration 1: Names and Input
-
+let hacker1 = "Phillip";
+console.log(`The driver's name is ${hacker1}`);
+let hacker2 = "Clara";
+console.log(`The navigator's name is ${hacker2}`);
 
 // Iteration 2: Conditionals
-
+if (hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`);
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,78 @@
...
+  }
+
+// Iteration 3: Loops
+let nameWithSpaces1 = "";
+
+for (let i=0; i < hacker1.length; i++){
+  nameWithSpaces1 += hacker1[i] + " ";
+}
+
+console.log(nameWithSpaces1.slice(0,13).toUpperCase());
```
fenrew: This 13 number that is added is hard coding. If you have a name that is 10 characters or 20 characters then this will not work. It only specifically works for names with 13 characters. Try to make it more dynamic. Hint: .length

js/index.js
```diff
@@ -1,7 +1,78 @@
...
+
+let nameWithSpaces2 = "";
+
+for (let j=hacker2.length-1; j >= 0; j--) {
+  nameWithSpaces2 += hacker2[j];
+}
+
+console.log(nameWithSpaces2);
+
+if (hacker2[0] < hacker1[0]) {
```
fenrew: How about: Andrè and Alfonso? These two names will both trigger the console.log("What?! You both have the same name?")

Instead you can solve this with localeCompare() to prevent this bug

PR #778, opened 2020-03-23
BER-FTWD-Agustina-Timur

HeiseMo:


js/index.js
```diff
@@ -1,7 +1,65 @@
 // Iteration 1: Names and Input
 
+let hacker1 = "Tim";
+console.log("The driver's name is " + hacker1)
```
fenrew: You can also use the backticks and ${hacker1}

js/index.js
```diff
@@ -1,7 +1,65 @@
...
+  
+
+// Iteration 3: Loops
+
+// 3.1
+
+let upperCaseDriverName = "";
+
+for (let i = 0; i < hacker1.length; i++) {
+  upperCaseDriverName = upperCaseDriverName + " " + hacker1[i].toUpperCase()
```
fenrew: Another way of writing this is:

  upperCaseDriverName +=  " " + hacker1[i].toUpperCase()

js/index.js
```diff
@@ -1,7 +1,65 @@
...
+
+let reversedNavigatorName = "";
+for (let i = hacker2.length - 1; i >=0; i--){
+  reversedNavigatorName += hacker2[i];
+}
+console.log(reversedNavigatorName)
+
+// 3.3
+
+if (hacker1.localeCompare(hacker2) === 1){
```
fenrew: Some browsers can return -2 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)
The best way to write this condition is either: 
hacker1.localeCompare(hacker2) > 0
hacker1.localeCompare(hacker2) < 0
hacker1.localeCompare(hacker2) === 0

js/index.js
```diff
@@ -1,7 +1,65 @@
...
+
+let words = loremIpsum.split(' ');
+console.log(words.length);
+
+let wordEt = 0;
+
+for (let i = 0; i < words.length; i++){
+    if (words[i] == 'et' || words[i] == 'et,' || words[i] == 'et.'){
```
fenrew: A bit of hardcoding here, but its fine since the challenge is difficult

PR #779, opened 2020-03-23
BER - FTWD - Maria and Benjamin 

Benchberlin:


js/index.js
```diff
@@ -1,7 +1,91 @@
...
+
+// Iteration 2 Conditionals 
+
+if (hacker1.length > hacker2.length) {
+    console.log(`the driver hast the longest name, it has ${hacker1.length} characters`);
+} else if (hacker1.length < hacker2.length) {
+    console.log(`the navigator hast the longest name, it has ${hacker2.length} characters`);
+} else if (hacker1.length === hacker2.length) {
+    console.log(`Wow, you both have equally long names ${hacker2.length} characters!`);
+}
```
fonsogms: nicely done. You could have also saved some code by using an `else{}`instead of `else if` in the last condition :)

js/index.js
```diff
@@ -1,7 +1,91 @@
...
+let hacker1Splitted = capitalizedHacker1.split(' ') ;
+
+console.log(hacker1Splitted) ;
+*/
+
+let result = '';
+
+for (i = 0; i < hacker1.length; i++) {
+    result += capitalizedHacker1[i] + ' ';
+}
```
fonsogms: nice job!

js/index.js
```diff
@@ -1,7 +1,91 @@
...
+for (i = hacker2.length - 1; i > -1; i--) {
+    result2 += hacker2[i];
+}
+
+console.log(result2);
+
+var arrNames = [hacker1, hacker2];
+
+arrNames.sort();
+
```
fonsogms: nice approach! You could also have used localeCompare() to sort the words on lexicographic order in case you are interested :)

PR #782, opened 2020-03-23
BER-FTWD-Daniela/Avtandil

daniela-kirilova:


js/index.js
```diff
@@ -1,7 +1,53 @@
...
+    `The driver has the longest name, it has ${hacker1.length} characters`
+  );
+} else if (hacker1.length < hacker2.length) {
+  console.log(
+    `it seems that the navigator has the longest name, it has ${hacker2.length} characters`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker1.length} characters!`
+  );
```
fonsogms: nice job!

js/index.js
```diff
@@ -1,7 +1,53 @@
...
+console.log(str);
+
+//3.2 Print all the characters of the navigator's name, in reverse order. i.e. "nhoJ"
+hacker2.split("");
+console.log(
+  hacker2
+    .split("")
+    .reverse()
+    .join("")
+);
```
fonsogms: nice approach! if you wanted to use a for loop you could also have used something like this
 ```
let str = "";
for (i = 0; i < hacker1.length; i++) {
  str = str + " " + hacker1[i].toUpperCase();
}
console.log(str);
```

File: js/index.js [hunk 3]
```diff
@@ -1,7 +1,53 @@
...
+//3.3 Depending on the lexicographic order of the strings, print
+
+if (hacker1 < hacker2) {
+  console.log(`The ${hacker1} goes first`);
+} else if (hacker1 > hacker2) {
+  console.log(`Yo, ${hacker2} goes first definitely.`);
+} else {
+  console.log(`What?! You both have the same name?`);
+}
```
fonsogms: this is a nice aproach, but you also have to take in account that if both names starts with the same letter it will `console.log(What?! You both have the same name?)` and you only want that to happen only when they actually have the same name, so for that you can use localeCompare().
https://www.w3schools.com/jsref/jsref_localecompare.asp

PR #784, opened 2020-03-23
BER-FTWD-Jony-Paula

jonystn:


js/index.js
```diff
@@ -1,7 +1,60 @@
...
+    console.log(`The driver's name goes first.`);
+    break;
+  } else if (hacker1[i] < hacker2[i]) {
+    console.log(`Yo, the navigator goes first definitely.`);
+    break;
+  } else {
+    console.log(`What?! You both have the same name?`);
+    break;
+  }
+}
```
Jeff-Moraes: It seems that your relational operators ( > and < ) are in the wrong direction

js/index.js
```diff
@@ -1,7 +1,60 @@
...
+Mauris mattis ac mauris eu efficitur. Morbi quis dictum felis, a convallis mauris. Phasellus tincidunt eros quis tristique vulputate. Suspendisse eros nisi, accumsan ac consequat efficitur, pretium nec dolor. Maecenas sollicitudin iaculis ex a viverra. Curabitur placerat sagittis urna, a porttitor elit posuere sit amet. Aenean sit amet tellus imperdiet magna venenatis iaculis non nec tellus. Maecenas libero nulla, pulvinar eget ipsum non, egestas facilisis dolor. Vestibulum dictum convallis orci a aliquet. Donec vitae venenatis quam, vel aliquet turpis. Nam interdum eros odio. Nam sit amet leo lacinia, sodales urna nec, ultricies leo. In nec ante nunc. Donec placerat sapien at enim suscipit, eget blandit tortor porta.`;
+
+console.log(lorem.split('\n').length - 1 + lorem.split(' ').length);
```
Jeff-Moraes: you can use just
console.log(lorem.split(' ').length)

jonystn: We tried it, but we got the sum with 2 fewer words.

PR #794, opened 2020-03-25
[PAR WDFT] William H & Pauline D ongoing

williamhzo:


TatijanaRajcic:
Hey guys! Thanks for your lab! William are you French? In which case next time i'll answer you in French, but I was just not sure haha.

Here are some remarks on your lab:
- Iteration 1 : ok
- Iteration 2 : ok. If you want to make it even cleaner, you could write it inside a function, which takes two argument, and compare them.
- Iteration 3.1: it is not exactly what was asked, as for now you are displaying [ 'W', 'I', 'L', 'L', 'I', 'A', 'M' ] instead of W I L L I A M. You are missing a "join" at the end: `console.log(hacker1.toUpperCase().split('').join(" "));`
- Iteration 3.2: OK!
- Iteration 3.3: This method works but compares only the first letter. Have a look at the function [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) that does the job for you, with every letter! 
- Bonus 1 : It's cool that you did the bonuses as well. If you want to go further with Bonus 1, here is a tip: It should work with every case of "et", even when it's followed by a "." or ",". Which is not the case right now since, in those cases, the "et" will be cut by your split function as a "et." or "et,"
- Bonus 2 : nice try but does not work with other palindroms that do not have a comma :D . What you co do is use regex to get rid of characters like "," "." etc. , and then only have the letters, join the whole thing, and then compare this full word, with its reverse. You'll know if it's a palindrome or not !

Nice job anyway!

PR #795, opened 2020-03-25
PAR WDFT Sam L Yuval B ongoing

samlfair:


ta-web-paris:
[Michel] OK. Below, my comments:
- 3.1 Your code adds a whitespace after the last letter (but shouldn't)
- Bonus "et": the second "var" for regexOutput is not necessary.
- Try to split too long lines into several lines (you can, e.g. after ||): it's way better than scrolling horizontally.
- Bonus palindrome: you should use regex again to match if each character is a letter.
- L108 : use "===" instead of "==": we want you to control the types of the compared elements, rather than letting == do it.

PR #798, opened 2020-03-25
[PAR-FT-0320] - Mathias Gautier ongoing

saitam1990:
not with bonus iterations (yet)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ _You should make a PR (stands for Pull Request) as soon as you make any signific
 	  - `Wow, you both have equally long names, XX characters!`.
 
 ### Iteration 3: Loops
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Mathias !

Good job overall, a few small points :
- It. 1 => you forgot to print "The driver's name is XXXX" (but it's all right you did the other one so you know how to do it, just be careful to do every step !)

- It. 3 => You typed
>> console.log(`"${realName.toUpperCase()}"`)
You don't need to put `` or "" or ${} since you're going to print a value, so instead just type
>> console.log(realName.toUpperCase())

- It.3 => you forgot 3.3 ! I let you do it, tell me when you push your changes so I can review them !

Otherwise nice work, and don't hesitate if you have questions :)

Nina

PR #800, opened 2020-03-26
PAR WDFT March Vladimir Leloup | Divya Gopal ongoing

Vlp90:
Exercice done

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,64 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
 
+// Iteration 1: Names and Input
+let driver1 = 'Vladimir';
+
+console.log("The driver's name is " + driver1);
+
+let driver2 = 'Divya';
+
+console.log("The navigator's name is " + driver2);
 
 // Iteration 2: Conditionals
 
+if (driver1.length > driver2.length) {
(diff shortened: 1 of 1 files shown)


TatijanaRajcic:
Hey guys! Thanks for this nice work!
Here are some remarks:
- Iteration 1: OK! Don't hesitate to use back ticks ` ` and then call a variable with ${variable}, as it is going to save you some + :D 
- Iteration 2: OK! Since you know how to do it that way, you could go even further and put this code inside a function (that takes two arguments) to get some practice
- Iteration 3.1: almost perfect but you are adding an extra space after the last character. Try and find a way to fix it :) I see that you tried with join and split, nice!
- Iteration 3.2: nice! Nice use of split join and reverse. Oh and one tip: you don't have to use "charAt(i)" to access the i letter of one word. using the index as it was a array is enough (driver1[i])
- Iteration 3.3: nice use of localeCompare. You could have also compared the word themselves (driver1 < driver2, driver 1 > driver2 )
- Bonus 1: nice try but notice that with the code you wrote, you don't account for every et : you are missing "et," "et." . Try and fix this one :)
- Bonus 2: did you give it a try? There are some simple solutions actually, using regex, and comparing the sentence in a right order, and the reverse one.

If you want to give it a try don't hesitate to send me a message :) 

Good work!

PR #808, opened 2020-03-26
[PAR-WDTF-032020] – Rebecca Norén & Olivier Brulé 🌟 ongoing

rebeccanoren:
We've done all of the mandatory tasks :)

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,48 @@
 // Iteration 1: Names and Input
+const hacker1 = 'Rebecca'
+const hacker2 = 'Olivier'
 
+console.log("1.2 > The driver's name is " + hacker1 + '.')
+console.log("1.4 > The navigator's name is " + hacker2 + '.')
 
-// Iteration 2: Conditionals
+//      Iteration 2: Conditionals
+//      2.1 Depending on which name is longer
+for (var i = 1; i < hacker1.length; i++) {
+  //console.log(i);
+}
 
+for (var i = 0; i < hacker2.length; i++) {
+  //console.log(i);
+}
 
-// Iteration 3: Loops
+console.log(
+  '2.1 > Wow, you both have equally long names, ' + i + ' characters!'
+)
+
+//      Iteration 3: Loops
+//      3.1 Print all the characters of the driver's name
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Rebecca and Olivier :)

Welcome to JavaScript ! 
A few points to keep in mind, since it's a pair lab I'm not going to ask you to bring corrections to it but I need you to read these carefully for your future codes because you can improve some parts :
- **Iteration 1** : good
- **Iteration 2** : where did it go ? I have unfinished code here :( But i see that you started for loops, you don't need them, i would advise to use a "if / else if / else" here (the title is conditionals, that's a hint hehe) so you could have something looking like 
>>   if (driver.length > navigator.length){
    console.log(`The driver has the longest name, it has ${driver.length} characters.`)
  } else if (navigator.length > driver.length){
    console.log(`It seems that the navigator has the longest name, it has ${navigator.length} characters.`)
  } else{
    console.log(`Wow, you both have equally long names, ${driver.length} characters!`)
  }

- **Iteration 3.1** : You went very hard on yourselves with the "add last letter" thing, one easy fix is to just remove the " -1" at the end of the "hacker1.length" in your loop (-1 is not necessary since you specified you wanted `"<"` (so LESS than the length number, and not less or equal)
- **Iteration 3.2** : You used array / string methods, and they work ! But the 3-section is about loops, to make you practice. One solution here would be to use a for loop, starting from the end of the word, going to the beginning, like this :
>>  let spellLog="";
  for (let i=navigator.length-1; i>=0; i--){
    spellLog+=`${navigator[i]}`
  }
  console.log(spellLog);

- **Iteration 3.3** : same thing, your code works but you were expected to loop, it's just about wrapping your code in a loop like this :
>>   for(i=0; i<driver.length; i++){
    if(driver===navigator){
      break
    } else if(driver[i].localeCompare(navigator[i])<0){
      console.log("The driver's name goes first.");
      break
    } else if (driver[i].localeCompare(navigator[i])>0){
      console.log("Yo, the navigator goes first definitely.");
      break
    }
  } 

Otherwise nice job, you got the basics of JavaScript ! Please get back to me if something is not clear, I'm happy to help :)  Try to really read those and practice them this weekend !

PR #815, opened 2020-03-28
PAR WDFT 202003 - Anne Patana - Sebastien Dubois - Week1 Day3 Lab ongoing

apatana:


TatijanaRajcic:
Hey guys! Thanks for your nice work! Here are some remarks to make it even better:
- Iteration 1: OK
- Iteration 2: OK! Since you know how to write functions now, try and think of a way to write this all code inside a function. Here, it could be a function that takes two parameters (which will be hacker1 and hacker2 when you executes it :) ) and that will return one of those sentences. But it was not mandatory so it's fine :) 
- Iteration 3: one quick thing to make your code shorter: writing `result = result + ad;` is the same as writing `result += ad;` . Also, small mistake here, you are adding a space after the last letter, which does not need one. Also, it's nice that you used your own loop, but FYI you could have used two JS built in functions which are split (splits a string/sentence into a array of letters/words, depending of what you want to do) and join (joins array elements into a string/sentence, with the separator you want (" ", "-" ...)).
- Iteration 4: nice! Fyi you could have used split, reverse and join. Try it out :)
- Bonus 1 : (you are not console.logging anything so that's why we don't see any result in the console, but it's fine.I just added console.log(wordCount(paragAll)) ). Clever solution to use split with "et" and see how many parts of the sentence we have. Please be careful that your method does not work if there is a "et, " or a "et.". You could get rid of those easily with a regex for example.
- Bonus 2: you need to get rid of ponctuation, spaces, and capital letters :) otherwise using split reverse and join is a good idea!

Nice work guys ! Keep up the good work !

PR #830, opened 2020-03-30
[BCN-FT-032020] Hélène Stephan

HeleneStephan:


js/index.js
```diff
@@ -1,7 +1,124 @@
...
+    (i === j - 1 &&
+      phraseToCheck[i].toUpperCase() === phraseToCheck[j].toUpperCase())
+  )
+    console.log(`Your phrase "${phraseToCheck}" is a palindrome!`);
+  else if (phraseToCheck[i].toUpperCase() != phraseToCheck[j].toUpperCase()) {
+    console.log(`Your phrase "${phraseToCheck}" is not a palindrom`);
+    i = j + 1;
+  }
+  j--;
+}
```
tawebbcn: Could be simplified, but cleaning first the 
Doesn't work with "Amor, Roma" because of the comma. You could clean the phraseToCheck beforehand from any special characters, do the same with the reversed phraseToCheck, and then compare the 2.

PR #831, opened 2020-03-30
[BCN -FT -032020] Lucia Skudlik

luciaskudlik:
Exercise finished. Bonus exercises not completely correct though

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,105 @@
+console.log("I'm ready!");
+
+//Iteration 1
+
+//1.1
+let hacker1 = "Lucia";
+
+//1.2
+console.log(`The driver's name is ${hacker1}`);
+
+//1.3
+let hacker2 = "Sarah";
+
+//1.4
+console.log(`The navigator's name is ${hacker2}`);
+
+
+//Iteration 2
+
+//2.1
+if(hacker1.length>hacker2.length){
+console.log(`The driver has the longest name, it has ${hacker1.length} characters.`)
+} else if (hacker2.length>hacker1.length){
+console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters.`)
+} else {
+console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`)
+};
+
+//Iteration 3
+
+//3.1
+let str = "";
+
+for (let i = 0; i<hacker1.length; i++) {
+ str += hacker1[i].toUpperCase() + " " ;
+};
+console.log(str);
+
+//3.2 (why 'undefined' before name??)
+let str2 = "";
+for (let i = hacker2.length--; i>=0; i--) {
+ str2 += hacker2[i] + " ";
+};
+console.log(str2);
+
+//3.3
+if(hacker1.charAt(0)<hacker2.charAt(0)) {
+console.log("The driver's name goes first.")
+} else if (hacker2.charAt(0)<hacker1.charAt(0)) {
+console.log("Yo, the navigator goes first definitely.")
+} else {
+console.log("What?! You both have the same name?")
+};
+
+//Bonus 1
+
+let string = "Lorem ipsum dolor sit amet, consectetur."
+
+   //number of words in string
+
+let str8 = ""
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -0,0 +1,105 @@
...
+
+//Iteration 3
+
+//3.1
+let str = "";
+
+for (let i = 0; i<hacker1.length; i++) {
+ str += hacker1[i].toUpperCase() + " " ;
+};
+console.log(str);
```
tawebbcn: Very good! Now if you print the "str" between quotes, you'll see that you get a space at the end of the string => "L U C I A "
You could handle this case by checking if the character that you're iterating over is the last character of that string, and in that only case, not add any space at the end (cf. Uros' review)
Other solution (that do not involve loops): you could use the trim() method, that removes whitespace from both ends of a string

index.js
```diff
@@ -0,0 +1,105 @@
...
+
+//3.2 (why 'undefined' before name??)
```
tawebbcn: as Uros explained this morning, if you decrement the length at the end ("post-decrement"), it will return the value before decrementing (so that's why you get "undefined"). It would work if you would "pre-decrement" like so => let i = --hacker2.length //  returns the value after decrementing
or just => let i = hacker2.length-1

index.js
```diff
@@ -0,0 +1,105 @@
...
+let str2 = "";
+for (let i = hacker2.length--; i>=0; i--) {
+ str2 += hacker2[i] + " ";
```
tawebbcn: no space needed here ;)

index.js
```diff
@@ -0,0 +1,105 @@
...
+console.log(str2);
+
+//3.3
+if(hacker1.charAt(0)<hacker2.charAt(0)) {
+console.log("The driver's name goes first.")
+} else if (hacker2.charAt(0)<hacker1.charAt(0)) {
+console.log("Yo, the navigator goes first definitely.")
+} else {
+console.log("What?! You both have the same name?")
+};
```
tawebbcn: Only compares the first character of the string, so considers that both have the same names if they start with the same letter. You could do include those condition statements inside a for loop, and check every character (like Uros showed this morning during review, don't hesitate if you have any question about it).

index.js
```diff
@@ -0,0 +1,105 @@
...
+
+   //number of words in string
+
+let str8 = ""
+for(let i = 0; i<string.length; i++) {
+  if (string[i]===" ") {
+    str8 += 1 
+  }
+};
+console.log(str8);
```
tawebbcn: nice start. Here you count the spaces, take into account there is one space less than there are words (because no space before the first word). Also str8 should be initialised with a number value (0), else you increment a string inside the loop, instead of a number.

index.js
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+
+
+   //how often appears 'et'
+console.log(string.includes("et"));
```
tawebbcn: Here you could convert "string" into an array, iterate over each word, then check if the word equals "et" and increment a counter each time it does.

index.js
```diff
@@ -0,0 +1,105 @@
...
+console.log(secondPart);
+
+   //write second part backwards
+
+let str3 = ""
+for (let i = secondPart.length--; i>=0; i--){
+  str3 += secondPart[i]; 
+}
+
+console.log(str3);//why 'undefined' before string??
```
tawebbcn: here you get undefined for the same reason than above (post-decrement)

index.js
```diff
@@ -0,0 +1,105 @@
...
+
+if(phraseToCheck.slice(0,4)===str3) {
+  console.log("This variable is a Palindrome.")
+} else {
+  console.log("This variable is not a Palindrome.")
+}
```
tawebbcn: The overall logic is good, comparing the 2 halves of a string.
To make it dynamic (working with any string provided), you could check the following things:
- If string has even number of characters, Half it and reverse it
- If string has odd number of characters, let a character in the middle and compare the 2 ends 
- remove any special characters like spaces or commas

Another way would be to create a reverse string and compare it to the initial string.

PR #833, opened 2020-03-30
[BCN -FT -032020] Daniel Hamekasi

wangagat:


js/index.js
```diff
@@ -1,7 +1,114 @@
...
+}
+hacker1Str += hacker1[i].toUpperCase();
+console.log(hacker1Str);
+
+//3.2
+var hacker2Str = ``;
+for (var i = hacker2.length - 1; i >= 0; i--) {
+  hacker2Str += hacker2[i];
+}
+console.log(hacker2Str);
```
tawebbcn: Nicely done with a loop

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+
+//3.3 Didnt know if you expected us to use nested for loops? found this method instead, although maybe that wasnt the correct approach.
```
tawebbcn: Good that you found this method, you'll get to use it quite often.
You could have used a for loop also, nesting the conditions inside the loop and comparing each character like Uros showed this morning (don't hesitate if you want to review it).

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+
+//this works if I remove the paragraphs, as I am counting white spaces for words. not sure how to catch the 2 words after a paragraph break.
```
tawebbcn: When using backticks to wrap your string, this shouldn't be an issue...

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+var etCount = 0;
+var nextIndexOfTerm = lorem.indexOf(searchTerm); //first occurence of search term. returns -1 if none found.
+
+for (var i = 0; i < lorem.length; i++) {
+  if (lorem[i] === ` `) {
+    wordCount++;
+  }
+}
+wordCount++; //this is for the last word of the string which is not succeeded by a space.
+console.log(`Number of words in string: ${wordCount}`);
```
tawebbcn: Good solution
Another way would be, instead of checking the blank spaces (which may lead to some error), you could convert your string "lorem" to an array with the method .split(" "), and just print the length of the array.

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+
+//for the length of string, and if indexOf finds search term, count++ and update nextindex for search.
+for (var i = 0; i < lorem.length && nextIndexOfTerm !== -1; i++) {
+  etCount++;
+  nextIndexOfTerm = lorem.indexOf(searchTerm, nextIndexOfTerm + 1);
+}
+console.log(`Number of occurences of "${searchTerm}": ${etCount}`);
```
tawebbcn: Works fine :)
Another solution a bit more straight-forward would be to convert your string to an array and iterate over to see if any element of the array is equal to "et"

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+
+//same construction as above but for reverse Phrase
+for (var i = phraseToCheck.length - 1; i >= 0; i--) {
+  if (phraseToCheck[i] === " ") {
+    continue;
+  } else {
+    newReversePhrase += phraseToCheck[i].toLowerCase();
+  }
+}
+console.log(`Reverse phrase: ${newReversePhrase}`);
```
tawebbcn: Here instead of doing all the checks again, you could reverse the "newPhrase" that you just cleaned from spaces, saves you some code ;)

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+  } else {
+    currentLetterMatch = false;
+    console.log(`This string is not a palindrome.`);
+    break; //it works, but wasnt sure if a break inside an IF would break out of the parent loop. apparently it does :P
+  }
+}
+if (currentLetterMatch) {
+  //if the loop made it this far and currentMatch is still true, then it's a palindrome.
+  console.log(`This string is a palindrome!!!`);
+}
```
tawebbcn: No need to loop over each character, you could just compare if newPhrase is equal to newReversePhrase, in that case it's a palindrome ;)

js/index.js
```diff
@@ -1,7 +1,114 @@
...
+
+//Bonus 2
+
+var phraseToCheck = `Amor, Roma`;
+var newPhrase = ``;
+var newReversePhrase = "";
+
+for (var i = 0; i < phraseToCheck.length; i++) {
+  if (phraseToCheck[i] === " ") {
+    //this takes care of spaces, but symbols like ! or ? arent taken care of here. didnt want to have a bunch of checks for symbols.
```
tawebbcn: Here are two ways to remove special characters: 
- check if `phraseToCheck[i] >= 'A' && phraseToCheck[i] <= 'z' `(checks if the character is a latin capital letter or latin small letter)
- use the method replace() on the phraseToCheck to remove all occurrences of characters that are not letters inside the string

PR #837, opened 2020-03-30
lab-js-basics Alberto Pagoria WDFT March 20

alpagor:
Iteraciones principales completadas y intentos de realizar primer bonus

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,88 @@
-// Iteration 1: Names and Input
+// Iteration 1
 
+let hacker1 = "Alberto";
+let hacker2 = "Alberto";
 
-// Iteration 2: Conditionals
+//console.log(`Driver\'s name is: ${hacker1}`);
+//console.log(`Navigator\'s name is: ${hacker2}`);
 
+// Iteration 2
 
-// Iteration 3: Loops
+if (hacker1.length > hacker2.length) {
+  console.log(
+    `The driver has the longest name, it has ${hacker1.lenght} characters`
+  );
+} else if (hacker2.length > hacker1.length) {
+  console.log(
+    `It seems that the navigator has the longest name, it has ${hacker2.length} characters`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker1.length} characters!`
+  );
+}
+
+// Iteration 3
+
+//creamos un loop por cada carácter en el nombre de hacker1
+let upperHack = "";
+
+for (let i = 0; i < hacker1.length; i++) {
+  upperHack += hacker1[i].toUpperCase() + " ";
+}
+console.log(upperHack);
+
+// imprimimos el nombre del navegador al reverse usando un loop "negativo"
+
+let reverse = "";
+
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  reverse += hacker2[i];
+}
+
+console.log(reverse);
+
+// Alphabetical order
+
+if (hacker1.localeCompare(hacker2) < 0) {
+  console.log("The driver's name goes first.");
+} else if (hacker1.localeCompare(hacker2) > 0) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?");
+}
+
+// bonus 1
+
+let text =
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,88 @@
...
+//console.log(`Driver\'s name is: ${hacker1}`);
+//console.log(`Navigator\'s name is: ${hacker2}`);
 
+// Iteration 2
 
-// Iteration 3: Loops
\ No newline at end of file
+if (hacker1.length > hacker2.length) {
+  console.log(
+    `The driver has the longest name, it has ${hacker1.lenght} characters`
```
tawebbcn: small typo here on the "length" word

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+} else if (hacker2.length > hacker1.length) {
+  console.log(
+    `It seems that the navigator has the longest name, it has ${hacker2.length} characters`
+  );
+} else {
+  console.log(
+    `Wow, you both have equally long names, ${hacker1.length} characters!`
+  );
+}
+
```
tawebbcn: otherwise very good

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+// Iteration 3
+
+//creamos un loop por cada carácter en el nombre de hacker1
+let upperHack = "";
+
+for (let i = 0; i < hacker1.length; i++) {
+  upperHack += hacker1[i].toUpperCase() + " ";
+}
+console.log(upperHack);
```
tawebbcn: Very good! Now if you print the upperHack between quotes, you'll see that you get a space at the end of the string => "A L B E R T O "
You could handle this case by checking if the character that you're iterating over is the last character of that string, and in that only case, not add any space at the end (cf. Uros' review)
Other solution (that do not involve loops): you could use the trim() method, that removes whitespace from both ends of a string

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+
+// imprimimos el nombre del navegador al reverse usando un loop "negativo"
+
+let reverse = "";
+
+for (let i = hacker2.length - 1; i >= 0; i--) {
+  reverse += hacker2[i];
+}
+
+console.log(reverse);
```
tawebbcn: Well done for using a loop here

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+
+// Alphabetical order
+
+if (hacker1.localeCompare(hacker2) < 0) {
+  console.log("The driver's name goes first.");
+} else if (hacker1.localeCompare(hacker2) > 0) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?");
+}
```
tawebbcn: Good using the localeCompare() method.
You could also have used a loop here in order to check every character (like Uros showed this morning)

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+/*let wordCount = 0;
+  for (let i = 0; i < text.length; i++) {
+	if (text[i] != " ") {
+		wordCount += 1;
+  	}
+  }
+
+console.log(wordCount);*/
+
+/* sin el metodo .split me cuenta cada caracter de la función en lugar de cada palabra...no logro entender como puedo sortear el problema. He considerado que un 'string' es como unos caracteres sin espacio y he creado una condicion según la cual a cada iteración, sí el index del text es diferente al espacio en blanco (" ") me suma 1 a la variable wordCount...pero no...Con el método .split lo que hago, al final, es hacer que la variable text se "separa" en un array de cadenas y contar cuanto es largo el array resultante*/
```
tawebbcn: You're almost there! Right now you're iterating over a string so you check every character and you increment wordCount if it's not a space. So you get the number of characters.
To solve the problem without using the split method, you could do it the opposite way: increment wordCount only if it's a space. That way you get the total number of blank spaces that follow every word, and you just need to add 1 to that value (in order to count also the first word of the 'text' string that is not preceded by a space).

`let spaceCount = 0;
  for (let i = 0; i < text.length; i++) {
	if (text[i] === " ") {
		spaceCount += 1;
  	}
  }
console.log(spaceCount + 1);`

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+
+//console.log(wordCount.split(" ").length);
```
tawebbcn: The split() method would be the more effective solution indeed :))

js/index.js
```diff
@@ -1,7 +1,88 @@
...
+
+for (let i = 0; i < textArr.length; i++) {
+  if (textArr[i] === 'et') {
+    charCount++;
+  }
+}
+
+console.log(charCount);
+
+//el resultado es 3 pero debería ser 5...*/
```
tawebbcn: You did it right!
I just see 4 "et" in the "text" string. The result here is only 3, because the last one is actually "et," (with a comma), so you could for this case in the condition.
`if (textArr[i] === 'et' || textArr[i] === 'et,') {
    charCount++;
  }`

PR #863, opened 2020-05-07
AMS WEB FT MAY - Ana Durguz

annettdd:


mswidzinska:
* Technical Requirements (8/10)
* Creative Code (7/10)
* Code Quality (8/10)
Nice job. 
Lorem ipsum text is a description of what lorem ipsum is, not text per se.
You forgot to do exercise: 3.3.
Bonus 1 - you are not using "i" inside of the loop, the loop has no sense then.
Bonus 2 - you should use if statement in "reverseString" function

javascript.js
```diff
@@ -0,0 +1,70 @@
+var text = "Lorem Ipsum is" //and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
```
mswidzinska, line 1: not a Lorem ipsum text

PR #885, opened 2020-06-04
[MIAPT062020 - Sandra - Stefan] iteration 1- create console.log()

sandrabosk:


sandrabosk:
good job, keep working like this. ♥️

PR #941, opened 2020-06-11
[PAR-FT-0620] Axel Chauvin

Axel0310:
Bonuses algorithmes could be certainly improved.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,122 @@
 // Iteration 1: Names and Input
 
+var hacker1 = "Menhaj";
+console.log(`The driver name\'s is ${hacker1}`);
+
+var hacker2 = "Axel";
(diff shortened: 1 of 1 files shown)


TatijanaRajcic:
Hello Axel and Menhaj ! Thanks for your lab ! It is really good ! Some remarks:
- Iteration 3.1: you are adding an extra space after the last letter, which is useful. Do you have an idea on how to avoid that? :)
- Bonus 1 iteration 1: your solution works but there is a more elegant solution: using split for example :)
- Bonus 2 iteration 2: nicely done for the regex :)
- Bonus 2 iteration 3: there is indeed simpler solution hehe :) for example using a regex, join, split for one sentence, and then on top of that using reverse for the second one, and compare both chains of characters :) That way you cover all the cases
Nicely done anyway !

PR #943, opened 2020-06-13
done

Menhaj01:


TatijanaRajcic:
Hey ! Thanks for your lab! Some remarks:
- Iteration 1: OK
- Iteration 2: OK
- Iteration 3: OK but you are adding an extra space after the last letter which is not useful. Do you know how you can get ride of it?
- Bonus 1: i get an error while trying the launch the program on nodemon: `prompt is not defined` . It's because "prompt" is a function you can use in your browser (i.e. for example if you had this script loading along a HTML page) but not in nodemon. To test your code, simply store a string inside your variable palindrome.
Your regex is not correct: to get ride of spaces, ponctuation, etc. you should use this regex:  ` let reg = /[^a-z0-9]/gi;` . Have a look at this stackoverflow [post](https://stackoverflow.com/questions/388996/regex-for-javascript-to-allow-only-alphanumeric) or other forums. Now you are almost there. You'll simply need to use toLowerCase to get rid of capital letters.

```
let palindrome = "Amor, Roma";
function palindromeFunction(str) {
  let reg = /[^a-z0-9]/gi;
  let withoutAlphanumeric = str.replace(reg, "").toLowerCase();
  console.log(withoutAlphanumeric);

  let reverseWord = withoutAlphanumeric
    .split("")
    .reverse()
    .join("")
    .toLowerCase();
  console.log(reverseWord);

  if (reverseWord === withoutAlphanumeric) return true;
  return false;
}

```
And now it's working!

- wordCount: perfect!
- timesEt: be careful, you have an extra "et using split like this. What you could do is using split(" ") and then iterate over the array you get, and count the number of 'et'. Be careful you get ride of the punctation before doing that, otherwise you will not count the "et," "et." etc

Keep up the good work!

PR #970, opened 2020-07-02
[AMS - PT - June20] - Yuka&Kenouly 

Kenouly:


js/index.js
```diff
@@ -1,7 +1,57 @@
-// Iteration 1: Names and Input
+console.log("I'm ready!");
+//Iteration 1: Names and Input
+const driver = 'John'
+console.log(`The driver's name is ${driver}.`);
+const navigator = 'Joe'
+console.log(`The navigator's name is ${navigator}.`);
+
+//Iteration 2: Conditionals
+if ((driver.length) < (navigator.length)) {
```
guido-an, line 9: There is no need to wrap the variables between parenthesis:
`if(driver.length < navigator.length)`

js/index.js
```diff
@@ -1,7 +1,57 @@
...
+    console.log(`It seems that the navigator has the longest name, it has  ${navigator.length} characters.`);
+}
+else if ((driver.length) > (navigator.length)) {
```
guido-an, line 12: Same as above

PR #973, opened 2020-07-04
First Homework assignment

Lisette-Sneha:


js/index.js
```diff
@@ -1,7 +1,86 @@
...
+console.log("I'm ready");
+const hacker1 = "Harry Potter"
+console.log("the drivers name is " +  hacker1);
+const hacker2 = "Hermione Lodge"
+console.log("the navigators name is " + hacker2);
 
 // Iteration 2: Conditionals
 
+if (hacker1.length > hacker2.length) 
+{
```
guido-an, line 13: Don't go to a new line when opening the if statement: 
`if(<condition>){
}`

guido-an: Same for the following lines

js/index.js
```diff
@@ -1,7 +1,86 @@
...
+
+var driverGoesFirst = "The driver's name goes first.";
+var navigatorGoesFirstDef = "Yo, the navigator goes first definitely.";
+var whatBothSameName = "What?! You both have the same name?";
+
+console.log(driverGoesFirst.localeCompare(navigatorGoesFirstDef, 'en', { sensitivity: 'base' }));
+
+console.log(navigatorGoesFirstDef.localeCompare (whatBothSameName, 'en', { sensitivity: 'base' } ));
+
+console.log
```
guido-an, line 50: console.log() should show some value on the console

js/index.js
```diff
@@ -1,7 +1,86 @@
...
+const result = []
+for (let i = 0; i < hacker1.length; i++) {
+  result.push(hacker1[i].toUpperCase())
+  result.push(' ')
+}
+console.log(result.join(''))
+
+function reverseString(str) {
+    
+    var splitString = hacker2.split("");
```
guido-an, line 33: You could still use the _str_ parameter in your function and then pass the variable hacker2 as an argument:
`function reverseString(str) {
    var splitString = str.split("");  
    var reverseArray = splitString.reverse(); 
    var joinArray = reverseArray.join(""); 
    return joinArray; 
}
reverseString(hacker2)`

Or remove the parameter _str_ that you are not using now

PR #986, opened 2020-07-27
[AMS-WEBDEV-JUL27] - Alex Anthopoulos

Kochasa:


reMRKableDev:
Nice work @Kochasa :thumbsup 
The only thing that would need looking over is the Bonus 1.b part. I think you want to run the ```.split()``` on the variable ```lorem``` instead of ```occurence``` in order to get the correct result.

PR #1026, opened 2020-08-26
[PAR-WDFT-248] Jorell & François 

JorellLangmar:
Please find the pull request with the iration from 1 to 3 as well as the bonus 1.

We experienced some problem though with the bonus 1 counting words because of the paragraphs.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,56 @@
+
 // Iteration 1: Names and Input
 
+var hacker1 = `Jorell`;
+//console.log (`The driver's name is ${hacker1}`)
 
+var hacker2 = `François`;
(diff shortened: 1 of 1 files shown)


N10T:
Hello Guys,

Try to commit for every iteration, "Iteration 1 | Done"....

**Iteration 3:**
Prefer to have your const initialisation outside of your loop

Reverse name:
Your reinitialize array twice 👎 :
var array = [];
var array = [];
The second one don't need the var again

- localeCompare 👍 

**BONUS**
- Regex 👍 you can have "or" in regex like "|" ;)       => /[ ]et[ .,]|[ .,]et[ .,]/
- Why not sum the frequency of "et" but good job
- isPalindrome is not working

good job 💯

PR #1028, opened 2020-08-26
[PAR-WDFT-248] Vincent Poum Raphaëlle Olivo 

RaphOL:
exercices faits ! Bonus 2 en cours

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -6,9 +6,22 @@ const hacker2 = "vincent";
 console.log("The driver's name is " + hacker1);
 console.log("The navigator's name is " + hacker2);
 
-
 // Iteration 2: Conditionals
 
+let drLgth = hacker1.length;
+// console.log(drLgth);
+let nvLgth = hacker2.length;
+// console.log(nvLgth);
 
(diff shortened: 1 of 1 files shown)


Fanny-Sbn:
Hello Ironhackers,

Nice for commiting at each iteration done. 

Iteration 1 👍 

Iteration 2 Instead of using 3 times if use if, else if and else 

Iteration 3 

3.1 It’s working but you have an extra space in front of all the letters, you need to add a condition to check if the it’s the first letter, if yes add the letter without the space 😉

3.2 👍 

3.3 Your program isn’t working… With hacker1[0] or hacker2[0] you’re checking only the first letters of the names, but what if they have the same first letter but not the same one for the second or the third, etc… You need to check all the letters and the length of the name too. Have a look to the method localeCompare, it will make your life super easy. 
Bonus 1 You needed to store 3 paragraphs in one variable, not in 3 different ones. You’ll see that it’s a bit more complicated. I let you try to find the solution 😉

Bonus 2 Super nice ! 

Good job ! 👍

PR #1030, opened 2020-08-26
[PAR-WDFT-248] Frederico Ratsimandresy

Ratsimandresy:
Didn't finish all bonus :(

### DIFF
--- a/js/dist/index.dev.js
+++ b/js/dist/index.dev.js
@@ -69,8 +69,48 @@ var lexicography = function lexicography(name1, name2) {
   }
 };
 
-console.log(lexicography(hacker1, hacker2)); // Bonus
+console.log(lexicography(hacker1, hacker2));
+console.log("--------------------------------"); // Bonus
 
-var phraseToCheck = "madam";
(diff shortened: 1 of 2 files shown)


N10T:
Hello Rico,

Cool you commit but try to be more explicit, "Iteration 1 | Done", "Feature sort alphabetically | Done"

- clean console 👍 

**Iteration 3**
- your function hackerNameToUppercase, leaving a space at the end of the word 👎  don't try to fix it with trim, fix your loop.
- cool reverse function loop 👍 

**3.1**
- The 2 options are working 👍 but option A is using a loop 🥇 

**3.3**
- lexicography... prefer to use a function like localeCompare

**Bonus 1**
- the regex is working 👍 but tries to understand it
- line 97 - don't forget to initialise with a let your variable in your for loop"for (let word of loremWords)"

**Bonus 2**
palindromCheck is not working for every palindrome 👎  "step on no pets" for example
palindromCheck is a variable with a callback function assign 👍 cool

good job 👍

Ratsimandresy:
Hello Andy,
Ok thanks for the feedback. Honestly, Luca helped me a lot with the regex
but i got everything so far.
Will do the changes!
Thank you

PR #1031, opened 2020-08-26
[PAR-WDFT-248] FABIEN MARCHAND AND NICOLAS ENAULT

nenault:


N10T:
Hey Coders,

- Frequent explicit commits 👍 
 
**iteration 1** 
- easy 👍 

**iteration 2**
- easy 👍 

**iteration 3**
- newDriversName finish with one space, try to fix your loop ;) maybe a condition

**3.3**
- why slugify if you compare 2 strings at the end?
- your 2 first keys of map object are the same, key has to be unique in an object
- prefer to use localeCompare, to compare that kind of strings

**bonus 1**
- **line 89: **what the function trim is doing, you don't need it 
- you are counting all "et" inside words too, we are looking for the "et" word

**bonus 2**
- function "reverseStringPal", why not reuse "reverseString"
- line 112, you can add the space in your 1st regex => in this case you don't need "removeSpace" and "cleanString"
var regex = /[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~ ]/g;
// removing special char by nothing
function cleanString(string) {
  return string.replace(regex, "");
}

function checkPalindrome(string) {
  let cleanedPhraseToCheck = cleanString(phraseToCheck);
  let cleanedreversedPhrase = cleanString(reversedPhrase);

Thanks for the nice presentation of the console log 💯 Good job but I'm not sure you understood the regex 🥇

PR #1032, opened 2020-08-26
[PAR-WDFT-248] Luca Marongiu

AyloSrd:


N10T:
Hello Luca,

Try to commit for every iteration: "Iteration 1 | Done", "Iteration 3 | 1 Done"...


**iteration 1**
- 👍 
-  nice console.log cleaned with line

**iteration 2**
-  👍 


**iteration 3**
- **3.1**  .trim() is solving a problem of conditions in your loop, but it's working
- **3.2** ok
- **3.3** ok (prefer to use a function like localeCompare)

**bonus 1**
- nice loop for this function homemadeSplit 👍 
- etCount work but if you have a "et," it's not working anymore. but ok.


**bonus 2**
- checkPalindromity 👍 
- be careful with this syntax "condition && action"

good job 👍

PR #1038, opened 2020-08-27
[PAR-WDFT-248] Bouziane and Marina

Bouziane95:


Fanny-Sbn:
Hello coders, 

Try to commit for every iteration: "Iteration 1 | Done", "Iteration 3 | 1 Done"...

iteration 1 & 2 👍 Try to use the backticks sometime ${variable}
 
iteration 3
3.1 - line 23: the driver finishes with a space "B_O_U_Z_I_A_N_E_" and just for you to know hacker1.charAt(i) is equal to hacker1[i]
3.2 : 👍
3.3 : Your program isn't working, you're comparing the first letter of both names but if the letter is the same what happens next ? In your program if the first letter of both name are the same you're printing that's they have the same names 26 times because you made a for loop on the length of the alphabet. 
Comparing strings is not conventional have a look to localeCompare() and try to find the solution !
 
Keep on going

PR #1107, opened 2020-10-12
[FTAMS102020] Nina van den Braak 

ninsajala:
Done iteration 1 and 2

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,18 @@
 // Iteration 1: Names and Input
+let hacker1 = "Sabela"
+//1.2 Print `"The driver's name is XXXX"`.
+console.log(`The driver's name is ${hacker1}`)
+//1.3 Create a variable `hacker2` with the navigator's name.
+let hacker2 = "Nina"
+//	1.4 Print `"The navigator's name is YYYY"`.
+console.log(`The navigator's name is ${hacker2} `)
 
 
 // Iteration 2: Conditionals
-
+if (hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`)
+} else if (hacker1.length < hacker2.length) {
+    console.log(`The navigator has the longest name, it has ${hacker1.length} characters.`)
+} else(console.log(`Wow, you both have equally long names, ${hacker1.length} characters! `))
 
 // Iteration 3: Loops
--- a/js/index.js
+++ b/js/index.js
@@ -19,11 +19,11 @@ if (hacker1.length > hacker2.length) {
 
 // Iteration 3: Loops
 //3.1
-hacker1 = hacker1.toUpperCase().split("").join(" ");
-
 for (let i = 0; i < hacker1.length; i++) {
-    console.log(hacker1.charAt(i))
+    console.log(hacker1.charAt(i).toUpperCase())
+    console.log(` `)
 }
+
 //3.2
 for (let i = (hacker2.length - 1); i >= 0; i--) {
     console.log(hacker2.charAt(i))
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,7 +1,81 @@
...
-// Iteration 3: Loops
\ No newline at end of file
+//3.2
+for (let i = (hacker2.length - 1); i >= 0; i--) {
+    console.log(hacker2.charAt(i))
+}
+//3.3
+switch (hacker1.localeCompare(hacker2)) {
+    case 1:
+        console.log(`The driver's name goes first.`);
```
wax-bear, line 34: Please be consistent in the use of semicolons. Either use them at the end of each statement or don't.

js/index.js
```diff
@@ -1,7 +1,81 @@
...
 
-// Iteration 3: Loops
\ No newline at end of file
+//3.2
+for (let i = (hacker2.length - 1); i >= 0; i--) {
+    console.log(hacker2.charAt(i))
+}
+//3.3
+switch (hacker1.localeCompare(hacker2)) {
+    case 1:
```
wax-bear, line 33: This output looks good. Please check the MDN documentation. "2" and "-2" are possible outcomes as well. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+} else if (hacker1.length < hacker2.length) {
+    console.log(`The navigator has the longest name, it has ${hacker1.length} characters.`)
+} else {
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters! `)
+}
 
+// Iteration 3: Loops
+//3.1
+for (let i = 0; i < hacker1.length; i++) {
+    console.log(hacker1.charAt(i).toUpperCase())
```
LloydC, line 23: You don't need to use two console.log, you can just write `console.log(hacker1.charAt(i).toUpperCase())+ " "`

wax-bear, line 23: The name is printed in a separated "way", however, not on the same line (e.g. "S A B E L A"). Would it be possible to print the name in uppercase and divided by spaces on the same line?

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+    console.log(` `)
+}
 
-// Iteration 3: Loops
\ No newline at end of file
+//3.2
+for (let i = (hacker2.length - 1); i >= 0; i--) {
+    console.log(hacker2.charAt(i))
```
wax-bear, line 29: Similar question here: could "aniN" be printed on one line?

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+    return str.split(" ").length;
+}
+console.log(WordCount(loremIpsum));
+
+//et count
+function etCount(str) {
+    let et1 = str.split(' et ').length;
+    let et2 = str.split(' et.').length;
+    let et3 = str.split(' et,').length;
+    let etTotal = (et1 + et2 + et3)
```
wax-bear, line 60: Very nice, you thought about these cases!

js/index.js
```diff
@@ -1,7 +1,81 @@
...
+}
+console.log(etCount(loremIpsum))
+
+//Bonus 2
+
+function palindrome(str) {
+
+    let newStr = str.toLowerCase().split("");
+
+    for (var i = 0; i < (newStr.length) / 2; i++) {
```
wax-bear, line 73: Could you please explain the use of dividing by 2?

wax-bear (changes requested):
Great job, Nina! Please process the comments for a pull request approval.

js/index.js
ninsajala: Because both halves of the string should be the same, only iterating through half of it should be enough to see if it's a palindrome

PR #1108, opened 2020-10-12
[FTAMS102020] Irene Asensio

ihatepineapple:


js/index.js
```diff
@@ -1,7 +1,106 @@
...
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`)
+  }
 
+// Iteration 3: Loops
+/* 3.1 Print all the characters of the driver's name, separated by a space and [in capitals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase)
+  i.e. `"J O H N"` */
 
-// Iteration 3: Loops
\ No newline at end of file
+hacker1 = hacker1.toUpperCase().split("").join(" ");
```
wax-bear, line 21: Please be consistent in the use of semicolons. Either use them at the end of each statement or don't.

js/index.js
```diff
@@ -1,7 +1,106 @@
...
-// Iteration 3: Loops
\ No newline at end of file
+hacker1 = hacker1.toUpperCase().split("").join(" "); 
+console.log(hacker1);
+
+/* 3.2 Print all the characters of the navigator's name, in reverse order. 
+  i.e. `"nhoJ"`*/
+
+for (let i = (hacker1.length); i >= 0; i-- ){
+    console.log(hacker1.charAt(i))
```
wax-bear, line 28: The characters are printed reversed, however, could it be printed on one line rather than on many lines?

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+}
+
+/*3.3 Depending on the [lexicographic order](https://en.wikipedia.org/wiki/Lexicographical_order) of the strings, print: <br>
+    - `The driver's name goes first.` <br>
+    - `Yo, the navigator goes first definitely.` <br>
+    - `What?! You both have the same name?`*/
+
+switch(hacker1.localeCompare(hacker2)) {
+    case 1: console.log(`The driver's name goes first.`);
```
wax-bear, line 37: This output looks good. Please check the MDN documentation. "2" and "-2" are possible outcomes as well. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+
+Sed laoreet ante vel quam auctor efficitur. Curabitur non metus vitae quam sollicitudin aliquet. Nunc imperdiet mi tortor, maximus sollicitudin urna volutpat et. Nunc eget nunc malesuada, ultrices tellus et, mollis orci. Pellentesque dapibus diam felis, sit amet consectetur lacus mollis at. Nunc gravida purus eget nibh bibendum mollis. Nulla ultricies metus ac turpis laoreet posuere. Proin eu dolor vel augue venenatis congue. Pellentesque convallis lorem id odio mattis, non sodales leo hendrerit. Nunc at malesuada metus.`
+
+
+function WordCount(str) { 
+  return str.split(" ").length;
```
wax-bear, line 60: Elegant!

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+}
+
+console.log(WordCount(LoremStr));
+
+// Make your program count the number of times the Latin word [`et`] appears.
+
+function WordEtCount(str) { 
+    return str.split(`et`).length;
```
wax-bear, line 68: Nice! Question: does this solution look for the word or the substring "et"? ("Make your program count the number of times the Latin word et appears")

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+    var newStr= str.toLowerCase().split("");
+  
+    for(var i=0; i < (newStr.length)/2; i++){ 
+      if(newStr[i] == newStr[newStr.length-i-1]){ 
+       
+      } else 
+        return false; 
+    }
+    return true;
+  }
```
wax-bear, line 100: Please mind the indentation here.

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+  - "taco cat"
+  - "put it up"
+  - "Was it a car or a cat I saw?" and "No 'x' in Nixon".
+  
+  __Hint__: If you use Google to help you to find solution to this iteration, you might run into some solutions that use advanced string or array methods (such as _join()_, _reverse()_, etc.). However, try to apply the knowledge you currently have since you can build pretty nice solution with just using `for` loop, `if-else` statements with some `break` and `continue`... Just sayin' :smiley: */
+
+
+function palindrome(str) {
+
+    var newStr= str.toLowerCase().split("");
```
wax-bear, line 91: Is the split method necessary here?

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+  
+    for(var i=0; i < (newStr.length)/2; i++){
```
wax-bear, line 93: The division of newStr.length strikes as "tricky" to me. What if the number is an odd one? Could the entire string be evaluated?

js/index.js
```diff
@@ -1,7 +1,106 @@
...
+  
+  __Hint__: If you use Google to help you to find solution to this iteration, you might run into some solutions that use advanced string or array methods (such as _join()_, _reverse()_, etc.). However, try to apply the knowledge you currently have since you can build pretty nice solution with just using `for` loop, `if-else` statements with some `break` and `continue`... Just sayin' :smiley: */
+
+
+function palindrome(str) {
+
+    var newStr= str.toLowerCase().split("");
+  
+    for(var i=0; i < (newStr.length)/2; i++){ 
+      if(newStr[i] == newStr[newStr.length-i-1]){
```
wax-bear, line 94: Perhaps the bang operator could be used here, and then in the code block "false" can be returned.

wax-bear (changes requested):
Great job Irene! Please process the comments for a pull request approval. If you have any questions, please reach out to me.

js/index.js
ihatepineapple: good question, it was difficult to figure. At this point I think it looks for just _'et'_ so it could also be inside a word (substring).
If you look for _' et '_ (with spaces) you could also be missing sentences that end with **et.**

js/index.js
ihatepineapple: Thanks, I unified.  
Still not sure when semicolons are mandatory or not.
Is is "best practices"to use them or not?

js/index.js
ihatepineapple: yes, fixed

let hackerReverse = ""
for (let i = (hacker1.length-1); i >= 0; i--){
  hackerReverse += hacker1[i]
}
console.log(hackerReverse)

js/index.js
ihatepineapple: Just in case I added then these extra cases

js/index.js
ihatepineapple: I added the sapces to make sure it only searches for **et**, but it might break if **et** is at the end of a sentence. (which would be weird since **et** means and, that i'm aware of)

js/index.js
ihatepineapple: ok, tried fixing it. 
Still a bit confusing for me, but will get the grip of it eventually.

js/index.js
ihatepineapple: noted

js/index.js
ihatepineapple: i tried with odd strings and it kept working, because the middle one would still be used as breaking point, i would presume

js/index.js
ihatepineapple: made it work, but needed help from TA for that

PR #1125, opened 2020-10-12
[FTAMS102020] Roy-Arne Brussel

Roybrussel:
JavaScript Basic Algorithms completed


js/index.js
```diff
@@ -1,7 +1,44 @@
...
 
+let hacker1 = ("Jan");
+console.log(`The driver's name is ${hacker1}`);
+
+let hacker2 = ("Bert");
+console.log(`The navigator's name is ${hacker2}`);
 
 // Iteration 2: Conditionals
 
+if (hacker1.length == hacker2.length) {
```
wax-bear, line 11: Please be mindful about the difference between strict and "loose" comparison. In this case "==" works fine. However, by default, I would stick to "===". https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons?page=2&tab=Votes

js/index.js
```diff
@@ -1,7 +1,44 @@
...
+    console.log(`Wow, both have equally long names, ${hacker1.length} characters!`);
+  } else if (hacker1.length >= hacker2.length) {
```
wax-bear, line 13: Do you really want to check whether hacker1.length is longer _or equal_ to hacker2.length?

js/index.js
```diff
@@ -1,7 +1,44 @@
...
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters`);
+  } else if (hacker1.length <= hacker2.length) {
```
wax-bear, line 15: Does "<=" resonate with the console.log on the following line?

js/index.js
```diff
@@ -1,7 +1,44 @@
...
+console.log(nameOfDriver);
+
+// 3.2
+var reversedNameOfNavigator = "";
+for (let i = hacker2.length-1; i >= 0; i--) {
+  reversedNameOfNavigator += hacker2[i];
+}
+console.log(reversedNameOfNavigator);
+
+// 3.3
```
wax-bear, line 35: This output looks good. Please check the MDN documentation. "2" and "-2" are possible outcomes as well. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

js/index.js
```diff
@@ -1,7 +1,44 @@
...
+if (hacker2.localeCompare(hacker1) === 1) {
+    console.log("The driver's name goes first.");
+  } else if (hacker2.localeCompare(hacker1) === 0) {
+    console.log("What?! You both have the same name?");
+  } else if (hacker2.localeCompare(hacker1) === -1) {
+    console.log("Yo, the navigator goes first definitely.")
```
wax-bear, line 41: Please be consistent in using the semicolon or not at the end of a line.

wax-bear (changes requested):
Good job Roy-Arne! Please process the comments for a pull request approval. Please reach out if you happen to have a question.

PR #1151, opened 2020-10-19
[WDFTRM-20201019] - Gosia, Hugo, Marta

msastreharo:


js/index.js
```diff
@@ -1,7 +1,46 @@
-// Iteration 1: Names and Input
+let hacker1 = "Gosia";
+let hacker2 = "Hugo";
 
+for (let i=0; i < hacker2.length; i++) {
+  console.log(hacker2[i].toUpperCase());
+}
```
filipe-freire, line 6: In here, you could use the template literal syntax (the backticks) in order to print both names to the console!
Ex: console.log(`"The driver's name is ${insert variable with the drivers name here}"`

js/index.js
```diff
@@ -1,7 +1,46 @@
...
+  console.log(hacker2[i]);
+}
 
+if (hacker1[0] < hacker2[0]) {
+  console.log('The drivers name goes first');
+} else if (hacker2[0] < hacker1[0]) {
+  console.log('Yo, the navigator goes first definitely.')
+} else {
+  console.log('What?! You both have the same name?')
+};
```
filipe-freire, line 18: For this iteration, the goal is to compare both name's length! by checking if hacker1[0] < hacker2[0] you're comparing the two characters at index 0 in both names! Maybe you could make use of the length property? Ex: hacker.length

PR #1152, opened 2020-10-19
[WDFTRM-20201019] - Pauline and Marc

paulinemss:


filipe-freire (commented):
Great work Pauline & Marc! I believe that you have mastered the use of for loops in this lab, very nice job!

PR #1188, opened 2020-10-19
javascript-basic-algorithms

capineti:
I don´t know if I understood the 3.3 but I compared the first letters of each name and deppending  on the position in the alphabet I decided the order  .


MartaJank:
Please, write your name on the commit titles so we know who you are :)

PR #1254, opened 2020-11-11
[PAR-WDFT-911] Ann Lampa - Charlotte Oury

charlotte-on:


N10T:
Excellent commits !
Good job and nice bonuses !

PR #1255, opened 2020-11-11
[PAR-WDFT-911] Anojan Mylvaganam & Victorien Provenzano

anojanmyl:
We didn't do the Bonus 2 for now


N10T:
Hey ironhacker...
Your code is breaking
"prompt is not defined" and this is breaking all your code

AnojanWDFT2020:
I removed the prompt issue it should run properly now (the command prompt worked on REPL but I didn't know it's not working on visual code)

PR #1256, opened 2020-11-11
lab done Etouma Mina [PAR-WDFT-911] Etouma Bela Mina Samwail

Etouma-bela:
We didn't figure out how to did the 3.3 with a loop


N10T:
Hey good job, just have to finish it ;)

PR #1267, opened 2020-11-12
[PAR-WDFT-911] Lucas Souleyrasse - Kenton Cargile

LucasSlrs:
Lab almost done, last part is not working we tried in uppercases and lowercases.


N10T:
Hey good start 👍 ... localeCompare...

PR #1363, opened 2021-03-22
[WDFTAMS032021] -Jamal Nasser- Jeroen Van Bart

jamal-nasser:
First day lab [WDFTAMS032021] -Jamal Nasser- Jeroen Van Bart

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,103 @@
-// Iteration 1: Names and Input
+console.log("I'm ready");
+//Iteration 1: Names and Input
+
+let hacker1="Jeroen van Bart";
+
+console.log(`The drivers name is ${hacker1}`);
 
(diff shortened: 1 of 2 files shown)


jamal-nasser:
Fixed some iterations

js/index.js
```diff
@@ -1,7 +1,112 @@
...
+
+
```
reMRKableDev, line 83: Nice attempt with using a `function`, however, the idea here is to practice with the loops and conditionals you've learned.

reMRKableDev (commented):
Nice work @jamal-nasser 🚀 

Your iterations work as expected. Great to see that you and Jeroen could collaborate well on this. I left a comment on the bonus iteration. Check it out

PR #1376, opened 2021-03-22
[FTAMS032021] Andrea Sforza - Igor Godinho Figueiredo

arsforza:
Completed both the mandatory and the bonus iterations.


reMRKableDev (commented):
Good work @arsforza 🚀

All your code is in check and in order. Kudos to you and Andrea for tackling the bonus section as well

PR #1389, opened 2021-03-22
done

silrodcar:


js/index.js
```diff
@@ -1,7 +1,39 @@
...
+let hacker1 = "Mayhem";
+console.log(`The driver's name is ${hacker1}`)
 
+let hacker2 = "Chai Latte"
+console.log (`The navigator's name is ${hacker2}`)
 
 // Iteration 2: Conditionals
 
+if (hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.lenght} characters`);}
```
reMRKableDev, line 11: Update this line to "length" instead of "lenght". Otherwise, the return value here will be `undefined` instead of the actual value

```suggestion
    console.log(`The driver has the longest name, it has ${hacker1.length} characters`);} 
```

reMRKableDev (changes requested):
Nice work @silrodcar 🚀 

All your required iterations are done and seem to function well. I've left a comment on a change you need to make. Check it out and make the necessary changes!

PR #1713, opened 2021-11-04
done

joaonacho:
Vitor Afonso Joao Nacho - Lisbon PT WebDev

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,7 +1,51 @@
 // Iteration 1: Names and Input
-// 
+const hacker1 = "Joao Nacho";
+console.log(`The driver's name is ${hacker1}`);
 
+const hacker2 = "Vitor Afonso";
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Hi guys, nice job.
Just a quick note on your final solution, it is always missing a word, since it's looking for the spaces but then it always misses the last word since there is no last space.
A simple solution would be simply to use .split() to turn the string into an array my the spaces, .split(" ") and then use .length to check the size
Also, another note, always test your solution with different variables - That was a huge chunk of text, so you could've used a string with only 3 words to test and then it would be easier to see if the solution was correct

Overall good work!

joaonacho:
Thanks @carlarsmendes for the feedback. After finishing the exercise, i thought of that situation, because the last word will not have a space after. But next time i will use the .length just to be sure that it is working just fine.

Thank you.

PR #1714, opened 2021-11-06
Javascript basic algorithms - WDPT Nov - Diogo and Michael

micfol:
Lab completed.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -31,8 +31,12 @@ for (let i = hacker2.length-1; i >= 0; i--) {
 }
 console.log(navNameReverse);
 
-if ("a" > "b"){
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Interesting and good lab solutions. Good usage of localeCompare even though we didn't talk about it in class - I didn't want to confuse you, but it's great that you used it.
About the method for countWords -> Clearly a regex expressions way which is out of scope for this bootcamp, but I encourage you to look at it at the end of the bootcamp :)
Good usage of switch and good naming of variables
Extra points for the ternary operator -> But be careful when using it, make sure it doesn't decrease the readability
Overall great job guys!

PR #1715, opened 2021-11-06
WDPT - Ana and Corentin

AnaModesto21:


carlarsmendes:
Hi Ana, Corentin, I know this was a difficult lab so don't worry too much, the bonus were not important 
You had some good solutions and the bonus are not that important
Don't forget to always write your ; at the end of your sentence: console.log();
I hope today's lesson and labs helped you understanding some concepts better, if not let me know what you're still struggling with

PR #1783, opened 2022-01-10
WDFT Remote Helena Saborit

Helsinky91:
Already pulling a request after iteration1 so I don't forget later 👯  I'll keep working on the other iterations

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,5 +1,9 @@
 // Iteration 1: Names and Input
+let hacker1 = "Helena";
+console.log("The driver's name is" + hacker1);
 
+let hacker2 = "X";
+console.log("The navigator's name is" + hacker2);
 // Iteration 2: Conditionals
 
 


js/index.js
```diff
@@ -1,6 +1,36 @@
...
 
-// Iteration 3: Loops
+// iteration 2: conditionals
+if (hacker1.length > hacker2.length) {
+  console.log(`The driver has the longest name, it has ${hacker1.length} characters`); 
+} else if (hacker1.length < hacker2.length) {
+  console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`);  
+} else if (hacker1.length == hacker2.length) {
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
```
TA-Remote, line 16: good job, just so you know: you could also write if, else if, else like this:
...
else {
  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
}

js/index.js
```diff
@@ -1,6 +1,36 @@
...
+let myArray = [hacker1, hacker2]
+myArray.sort();
+
+if (myArray[0] == hacker1) {
+  console.log("The driver's name goes first.");
+} else if (myArray[0] == hacker2) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?")
+}
```
TA-Remote, line 34: 3.3
actual an excellent solution, but as you already know, it should be solved differently here, which is not a problem!!!

js/index.js
```diff
@@ -1,6 +1,36 @@
...
+  console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`);  
+} else if (hacker1.length == hacker2.length) {
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
+
+// iteration 3: loops
+//3.1
+let spaces= hacker1.toUpperCase(); 
+console.log(` ${spaces.charAt(0)} ${spaces.charAt(1)} ${spaces.charAt(2)} ${spaces.charAt(3)} ${spaces.charAt(4)} ${spaces.charAt(5)} `);
+//3.2
```
TA-Remote, line 22: That's what you and Luis were talking about this morning. So no problem that you got it wrong and great that you didn't just copy the other solution.

PR #1791, opened 2022-01-10
WDFullTimeJan22 Bart and Marten

ZandMa:


js/index.js
```diff
@@ -1,6 +1,27 @@
...
+if (hacker1.length > hacker2.length){
+    console.log("The driver has the longest name, it has " + hacker1.length + " characters.");
+    console.log("the drivers name goes first")
+  } else if (hacker2.length > hacker1.length) {
+    console.log("It seems that the navigator has the longest name, it has " + hacker2.length + " characters.");
+    console.log("Yo, the navigator goes first definitely.");
+  } else {
+    console.log("Wow, you both have equally long names, " + hacker1.length + " characters!");
+    console.log("What?! You both have the same name?");
+  }
```
TA-Remote, line 19: Task 1 and 2 are good

TA-Remote: you didn't do 3.3, it should be the alphabetical order

js/index.js
```diff
@@ -1,6 +1,27 @@
...
+   
+  var resultsDriver = hacker1
+  console.log(resultsDriver.split('').join(' ').toUpperCase());
+  
+  var resultsNavigator = hacker2
+   console.log(resultsNavigator.split("").reverse().join(""));
```
TA-Remote, line 26: you didn't do 3.3, it should be the alphabetical order

js/index.js
```diff
@@ -1,6 +1,27 @@
...
+    console.log("It seems that the navigator has the longest name, it has " + hacker2.length + " characters.");
+    console.log("Yo, the navigator goes first definitely.");
+  } else {
+    console.log("Wow, you both have equally long names, " + hacker1.length + " characters!");
+    console.log("What?! You both have the same name?");
+  }
+   
+  var resultsDriver = hacker1
+  console.log(resultsDriver.split('').join(' ').toUpperCase());
+
```
TA-Remote, line 23: var is not really used nowadays. And you could have just written hacker1 instead of resultsDriver
or resultsDriver = hacker1.split('').join(' ').toUpperCase()

This is a great solution for later! But actually you should use a for loop here

js/index.js
TA-Remote, line 26: same for this one. But all those methods will be very useful later

TA-Remote: We upload solutions in the Student Portal. Feel free to check them out. If you have any questions, please feel free to contact us at any time. The fact that you are good with the methods is already a good sign either way.

PR #1793, opened 2022-01-10
Nico 

NicoTollenaar:
Initial commit without bonus


js/index.js
```diff
@@ -1,7 +1,205 @@
...
 // Iteration 2: Conditionals
 
+if (hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`);
+} else if (hacker2.length > hacker1.length) {
+    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`);
+} else if (hacker1.length === hacker2.length) {
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters`)
+}
+
```
TA-Remote, line 22: good job, another possibility would have been to use if, else if and else. So like this:
if (hacker1.length > hacker2.length) {
    console.log(`The driver has the longest name, it has ${hacker1.length} characters.`);
} else if (hacker2.length > hacker1.length) {
    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`);
} else {
    console.log(`Wow, you both have equally long names, ${hacker1.length} characters`)
}

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+    let newString = "";
+    for (let i = 0; i < name.length; i++) {
+        if (name[i+1]) {
+            newString += `${nameInUpperCase[i]} `; 
+        } else {
+            newString +=`${nameInUpperCase[i]}`;
+        }
+    }
+    console.log("New string is: ", newString);
+    return newString;
```
TA-Remote, line 37: Great solution! Very good that you wrote a function for it. (Instead of just writing the for loop etc.) I also like that you give the functions a meaningful name.

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+}
+
+printInCapitalsWithSpaces(hacker1);
+printInCapitalsWithSpaces(hacker2);
+
```
TA-Remote, line 42: here I first missed the task 3.2, but since you wrote almost the same function for the second bonus task, this is not a problem

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+        case navigator:
+            console.log("Yo, the navigator goes first definitely.");
+            break;
+        case "equal":
+            console.log("What?! You both have the same name?");
+    }
+}
+
+commentOnLexicographicalOrder(hacker1, hacker2);
+
```
TA-Remote, line 82: Good solution, but it would have been a little easier. If you are interested, just have a look at the solution we will upload in the Student Portal

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+phraseToCheck("Amor, Roma");
+phraseToCheck("race car");
+phraseToCheck("stack cats");
+phraseToCheck("step on no pets");
+phraseToCheck("taco cat");
+phraseToCheck("put it up");
+phraseToCheck("Was it a car or a cat I saw?");
+phraseToCheck("No 'x' in Nixon");
+phraseToCheck("Random other phrase");
+
```
TA-Remote, line 198: great!

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+        if (string[i]= " ") {
+            numberOfWhiteSpaces++;
+        }
+
+    }
+    
+    const numberOfWords = numberOfWhiteSpaces - 1;
+    console.log("Number of words of Lorem text: ", numberOfWords);
+    return numberOfWords;
+}
```
TA-Remote, line 108: Here you have a logic mistake. You count the letters instead of the words. And it's +1 when it comes to words compared to spaces.

js/index.js
```diff
@@ -1,7 +1,205 @@
...
+
+countNumberOfWords(lorem);
+
+function countNumberOfEts(string) {
+    
+    let index;
+    let slicedString;
```
TA-Remote, line 115: you don't use "slicedString";

PR #1794, opened 2022-01-10
[REMOTE-FT-012022] Zena Bos

zenabos:
We did the three iterations together, the two bonus we did separately.


js/index.js
```diff
@@ -1,6 +1,91 @@
...
+
+console.log(resultString.toUpperCase());
+
+let reverseString = '';
+
+for (let j = hacker2.length - 1; j >= 0; j--) {
+  reverseString = reverseString.concat(hacker2[j], ' ');
+}
+
+console.log(reverseString);
```
TA-Remote, line 45: good job! for (
let i = 0; i < hacker1.length; i++) {
  resultString = resultString.concat(hacker1[i], ' ').toUpperCase();
}

console.log(resultString);

should also work

js/index.js
```diff
@@ -1,6 +1,91 @@
...
+function countEt (lorumIpsum) {
+  let count = 0;
+  for (let i = 0; i < lorumIpsum.length; i++)
+  if((lorumIpsum[i] == " ") && (lorumIpsum[i+1] == "e") && (lorumIpsum[i+2] == "t") && (lorumIpsum[i+3] == " " || "." || ",")) {
+      count += 1;
+    }
+  console.log(count);
+}
+
+countEt(lorumIpsum)
```
TA-Remote, line 75: pretty good!

js/index.js
```diff
@@ -1,6 +1,91 @@
...
+      for (let i = 0; i < phraseLength / 2; i++) {
+        if (phraseToCheck[i] !== phraseToCheck[phraseLength - 1 - i]) {
+          console.log('It is not a palindrome')
+        } else {
+            console.log('It is a palindrome')
+        }
+      }
+  }  
+  
+  checkPalindrome("value")
```
TA-Remote, line 91: Here it is best to look again at the solutions we will send you, but the rest is excellent

zenabos: Yeah I would love to see the solutions, because I got really stuck at the end. I've tried to work with break and continue, but then I got errors. :(

PR #1805, opened 2022-01-11
REMOTE-FT-012022- Tobias Mulatsch

Rospini:


js/index.js
```diff
@@ -1,6 +1,60 @@
...
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`)
+}
+else if (hacker1.length < hacker2.length){
+  console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters.`)
+
+}
+else if (hacker1.length > hacker2.length){
+  console.log(`The driver has the longest name, it has ${hacker1.length}characters.`)
+}
+
```
TA-Remote, line 19: good job, but here you could also do just if, else if, else (instead of if, else if, else if)

js/index.js
```diff
@@ -1,6 +1,60 @@
...
+ 
+    
+let driverName = hacker1.split("").join(" ");
```
TA-Remote, line 22: Super work, so it is a super solution later. Actually it was meant that you use a for loop here (for now)

TA-Remote: let driverCapitals = ""

  let upperCaseLetter = hacker1[i].toUpperCase()
  
  // if it is the first letter to be added, then add it without the space
  if (i === 0) {
    driverCapitals = driverCapitals + upperCaseLetter
  } else {
    driverCapitals = driverCapitals + " " + upperCaseLetter
  }
  
}

console.log(driverCapitals)

like this for example

js/index.js
```diff
@@ -1,6 +1,60 @@
...
+  let reverseHacker = " ";
+  for (var i = hacker2.length - 1; i >= 0; i--) {
+      reverseHacker += hacker2[i];
+  }
+  return reverseHacker
+  
+}
+let finalReverse = reverseNav(hacker2)
+console.log(finalReverse)
+
```
TA-Remote, line 37: perfect

js/index.js
```diff
@@ -1,6 +1,60 @@
...
+if (driverFirstLetter === navigatorFirstLetter ){
+  console.log("What?! You both have the same name?")
+}
+else if(driverFirstLetter < navigatorFirstLetter ){
+  console.log("The driver's name goes first")
+}
+else if(driverFirstLetter > navigatorFirstLetter ){
+  console.log("Yo, the navigator goes first definitely.")
+}
+
```
TA-Remote, line 50: the solution is kind of right, but it will only compare the first letter. So if we have Anna and Anne we have a problem.

TA-Remote: you can check the solution later in the student portal. you  need a loop that runs for every character and compares them.

js/index.js
```diff
@@ -1,6 +1,60 @@
...
+Duis quis dolor lacus. Aliquam erat volutpat. Duis dapibus lacus diam, elementum facilisis libero aliquet eget. Morbi in felis vel sem varius pretium. Sed aliquam vitae mauris ut sodales. Curabitur maximus ex urna. Sed a hendrerit mi. Nunc feugiat justo placerat maximus molestie.
+
+Aenean magna nibh, vehicula non sodales quis, vestibulum lacinia velit. Donec id turpis nibh. Sed eget justo maximus ligula mollis auctor. Integer ligula diam, faucibus ac sapien sed, commodo euismod nulla. Duis quis commodo eros. Cras tempus, lacus in faucibus convallis, nunc erat eleifend risus, eu laoreet est diam sed turpis. Ut sagittis viverra ligula, a tempus nunc congue sed. Duis a ipsum sit amet ante accumsan rhoncus id at augue. Aenean eu ultrices felis, in fermentum nibh. Nulla facilisi. Lorem ipsum dolor sit amet, consectetur adipiscing elit.`
+
+function WordCount(lorem) { 
+  return lorem.split(" ").length;
+}
+console.log(WordCount(lorem));
```
TA-Remote, line 60: all in all a really good job!

PR #1844, opened 2022-01-15
Fernando Costa [lab-javascript-basic-algorithms] 

breogann:


breogann:
Hello! 🙋‍♂️

I solved Bonus 2, but not with the `if`, `else if` approach.

Could you check if that solution makes sense? Pretty much I tried to:

1. Reverse the word
2. Iterate over the original word, iterate over the palindrome backwards
3. Check if those indexes would match

But it still doesn't work. It goes into the first loop, but not the nested one. 

image

PR #1896, opened 2022-03-22
Lucas Perez WDFT rmtus Mar 2022 

LPerezDev:
Got through everything besides 3.3 last night. This morning we (Adam, my lab partner, and I) Tinkered with the conditional and got it to print what looks like the correct answers.


sunspla-sh:
Looks good.

PR #1921, opened 2022-03-28
Finished

infjay:


TA-Remote:
Hi Jawid! Overall really good job! Task 3 should be done with loops, but you still did a great job! If you want, you could compare task 3 again, we will upload the solutions in the Student Portal under Extra Resources. 

Great that you solved Bonus 1: 
I think you need to use backticks for the lorem ipsum text. 
And it must be like this:

`let etCount = text.split(" et").length` --> space before "et"

To make sure it counts only „et“ and not for example „laoreet“ as well.
You can also find the bonus tasks in the solutions if you want to take a look!

PR #1923, opened 2022-03-28
WDFT REMOTE MAR22 Fevi + Ihab

fevi-ro:


js/index.js
```diff
@@ -1,6 +1,37 @@
...
+} else if (hacker1.length < hacker2.length) {
+    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`);
+} else if (hacker1.length = hacker2.length) {
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
 
 // Iteration 3: Loops
+//3.1
+
+console.log(hacker1.toUpperCase());
```
TA-Remote, line 21: 3.1 iteration asks you to also separate letter with space ;)
So result should look like this `"J O N A T H A N"`

js/index.js
```diff
@@ -1,6 +1,37 @@
...
+//3.2
+
+splitName = ['a', 'm', 'a', 'n', 'd', 'a']
+let reverseName = splitName.reverse();
+hacker2 = reverseName.join('');
+console.log(hacker2);
+
```
TA-Remote, line 28: Always think that your code should be reusable, not attached to a specific value. In your case if I want to reverse name Peter, I have to create an array with all these letters. But the idea is that I only pass a string and the pre-written code does all the work for me! 

Another approach for this iteration could be using `for loop`

js/index.js
```diff
@@ -1,6 +1,37 @@
...
+//3.3
+
+if (hacker1.substring(0, 1) > hacker2.substring(0, 1)) {
+    console.log(`The driver's name goes first.`);
+} else if (hacker1.substring(0, 1) < hacker2.substring(0, 1)) {
+    console.log(`Yo, the navigator goes first definitely.`);
+} else if (hacker1.substring(0, 1) = hacker2.substring(0, 1)) {
+    console.log(`What?! You both have the same name?`);
+}
```
TA-Remote, line 37: If you were to compare letters, `a `should be less than `b `as it comes first in the alphabet 

Also quite important to remember 
- single `=` is used to assign a value
- double or triple  equal `==` / `===` is used to compare variables

TA-Remote (commented):
Good work both, couple of things to improve in iteration 3.
I left comments for you and will be uploading solution on student portal

PR #1925, opened 2022-03-28
[FTRMT032022] Alexander Nerz and Oscar Gonzalez

oscargonzalezdev:
Solve the Lab JavaScript Basic Algorithms
Exercise completed by: Alexander Nerz and Oscar Gonzalez.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,6 +1,69 @@
 // Iteration 1: Names and Input
-//
+
+let hacker1 = 'Alex'
+console.log(`The drivers name is ${hacker1}`);
+
+let hacker2 = 'Oscar'
+console.log(`The navigators name is ${hacker2}`);
+
+
 // Iteration 2: Conditionals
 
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,6 +1,69 @@
...
+
+let phraseToCheck = "race car";
+let phraseRev = phraseToCheck.split('').reverse().join('');
+console.log(phraseRev);
+
+if (phraseToCheck.replace(/\s/g, '') === phraseRev.replace(/\s/g, '')) {
+  console.log(`${phraseToCheck} is a Palindrome of ${phraseRev}`)
+} else {
+  console.log('its not a Palindrome');
+}
```
TA-Remote, line 69: Using functions already :ok_hand:
And regular expressions! Good thinking, just make sure you understand how they work if you use them ;)

oscargonzalezdev: Thanks for the feedback!

PR #1927, opened 2022-03-28
WDFT MARCH2022 Sepideh.Att+Sina.Kaabi

Sepidehatt:


sina-kaabi:
Sepideh completed the bonus questions. I completed questions 1 to 3 on my own and she did her own versions of 1 to 3 also but we worked on 3.1 together mainly as it was tricky for me. I submitted her lab to the portal but I will add my coding comments during lunch break to show what I tested that worked & what did not work etc.

PR #1929, opened 2022-03-28
Lab done with Thu

JeanDenisD:


TA-Remote:
Hi! Really good work, I have almost nothing to say!! We will upload solutions to Student Portal, if you want to compare some tasks again you can do it, but your solution is great.
Only thing: Bonus 2 with palindromes works but not with special characters. I answered questions about this in ask-the-class channel, if you want to understand how you could have solved this.

PR #1931, opened 2022-03-28
WD FT MAR22 Harry&Louise  

LouiseGabrielle0:


js/index.js
```diff
@@ -1,6 +1,144 @@
...
+  }
+}
+  backwardsWord.toLowerCase()
+  phraseToCheck.toLowerCase()
+
+if (spacelessWord.toLowerCase === backwardsWord.toLowerCase){
+  isPalindrome = true
+} else {
+  isPalindrome = false
+}
```
TA-Remote, line 140: Excellent work both!

The only note here. Currently in bonus2 solution you always get true. 

On line 136, inside if statement you forgot to add bracket.
- `toLowerCase() `is a method (function), so you have to 'call it' to use it. 
-  Otherwise you are just comparing two functions literally , but not the result of each function, not the string in lower case

PR #1933, opened 2022-03-29
WDFT 22March Diana and Alex

didicozaur:


js/index.js
```diff
@@ -1,6 +1,47 @@
...
+
+//3.3
+
+if (hacker1 < hacker2) {
+  console.log("The driver's name goes first.");
+} else if (hacker1 > hacker2) {
+  console.log("Yo, the navigator goes first definitely.");
+} else {
+  console.log("What?! You both have the same name?");
+}
```
TA-Remote, line 47: Excellent solutions in iteration3! Some advanced concepts there ;) 

Alternatively, think how could you have solved it using `for loop`

PR #1934, opened 2022-03-29
Js Algo Lab - Daniel + Marija - WDFT2022

marija-strah:


js/index.js
```diff
@@ -1,6 +1,85 @@
...
+// Bonus 2
+
+let phraseToCheck = "";
+
+function checkPalindrome(phraseToCheck){
+  let reversed = phraseToCheck.split("").reverse().join("");
+  return phraseToCheck === reversed
+}
+
+console.log(checkPalindrome('step on no pets'));
```
TA-Remote, line 85: I like that you have written a function for this. 
The function works but not with special characters (e.g. No 'x' in Nixon“). I answered questions about this in ask-the-class channel, if you want to understand how you could have solved this.

PR #1942, opened 2022-04-08
WDPT ABR22 Emmanuel Solis 

emmanuelsolis:
created two variables and printed their values


ta-web-mexpt2:
Buen bonus aplicado heeee 🔥

PR #1962, opened 2022-04-11
[FTLIS042022] Duarte Alves & Alexandre Alves

drttyy:


webLisbonTA (commented):
Good job on finishing all the iterations!

js/index.js
```diff
@@ -1,6 +1,39 @@
 // Iteration 1: Names and Input
-//
+var hacker1 = "Duarte";
+console.log(`The driver name is ${hacker1}` );
+var hacker2 = "Alexandre";
+console.log(`The navigator name is ${hacker2}`);
```
webLisbonTA, line 5: Try and use let/const instead of var. It still works but nowadays as a rule of thumb we try to use let and const

js/index.js
```diff
@@ -1,6 +1,39 @@
...
 // Iteration 2: Conditionals
-
+if(hacker1.length >= hacker2.length){
+    console.log(`The driver's as a longest name, it has ${driver1.length} characters.`)
+}else if( hacker1.length <= hacker2.length){
+    console.log(`It seems that the navigator has the longest name, it has ${hacker2.length} characters`)
+}else{
+    console.log(`Wow, you both have equaly long names, ${hacker1.length, hacker2.length} characters.`)
+}
```
webLisbonTA, line 13: Using the <= && >= will make your last condition redundant. You should do < / > in this case

js/index.js
```diff
@@ -1,6 +1,39 @@
...
+hacker1 = hacker1.toLowerCase;
+hacker2= hacker2.toLowerCase;
+
+if("Duarte".localeCompare("Alexandre") === -1 ){
+    console.log("The driver's name goes first.")
+}else if( "Duarte".localeCompare("Alexandre") === 1){
+    console.log("Yo, the navigator goes first definitely.")
+}else{
+    console.log("What?! You both have the same name?")
+}
```
webLisbonTA, line 39: For a better code organisation and readability, try and use variables instead of writing the whole name

PR #1964, opened 2022-04-11
[WDFTAPR22] Keven and Leticia 

kevenm4:


webLisbonTA (commented):
Great job, Keven! You solutions are great!
Try to do the bonus, if you have time! 

A hint - the count et is really similar to count Smiles from the prework

PR #2055, opened 2022-06-15
Tobias Conrad with Jack

maktooob:
DId the lab with Jack (not alex as the commit m- is saying :)


js/index.js
```diff
@@ -1,6 +1,41 @@
...
+}
+console.log(driver);
+let lexCompareNames = hacker1.localeCompare(hacker2);
+console.log(lexCompareNames)
+if (lexCompareNames === -1) {
+    console.log("Yo, the navigator goes first definitely.");
+}
+else if (lexCompareNames === 1) {
+    console.log("The driver's name goes first.");
+}
```
TA-Remote, line 38: You have it other way around.. if `localeCompare()` returns -1, it means hacker1 goes first 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

TA-Remote (commented):
Great job Tobias! 

Also, if we give you a chance to work in breakout rooms - that suggests you work at faster pace and definitely have time to do bonuses :) So next time make sure you go that extra mile - it will make huge difference

js/index.js
maktooob: Thank you for going over it. I actually did the bonuses, just forgot to push them - did now :)

PR #2063, opened 2022-06-15
Michael Oberst, together with Paul Tiry

miob1781:
I did this together with Paul Tiry.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,6 +1,101 @@
-// Iteration 1: Names and Input
-//
-// Iteration 2: Conditionals
+console.log("I'm ready!")
 
+// Iteration 1: Names and Input
+
+const hacker1 = "Paul"
+console.log("The driver's name is " + hacker1 + ".")
+const hacker2 = "Michael"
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,6 +1,101 @@
...
+            message = "The driver's name goes first."
+            break
+        } else {
+            if (shortestName.length !== longestName.length){
+                message = "The driver's name goes first." 
+            } else {
+                message = "The names are the same."
+            }        }
+    } 
+}
```
TA-Remote, line 78: In iteration 3.3, instead of repeating your code, you can onyl loop once, over the lenght of longst word .

And add additional check, for example
```
hacker1[i] < hacker2[i] || hacker1[i] === undefined
```
// second condition will check if the letter of the driver is lower (goes first) or doesn't exist.

This way if there's no more letter at that index -> you  have the answer, shorter word will go first

TA-Remote (commented):
Great work Michael, keep it up!

If you have extra time, make sure to go that extra mile and work on bonuses, the more time you put in coding, the bigger difference you'll see in long term ;)

PR #2065, opened 2022-06-15
Paul Tiry [WDFT JUN2022] with Michael Oberst

ptiry02:
Hi, this is my lab which I started with Michael. They may be different because I formatted my code to my liking and also I did the bonus exercises by myself.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -9,6 +9,7 @@ console.log(`The driver's name is ${hacker1}`);
 const hacker2 = "Michael";
 
 console.log(`The navigator's name is ${hacker2}`);
+console.log();
 
 // Iteration 2: Conditionals
 
@@ -25,6 +26,7 @@ if (hacker1.length > hacker2.length) {
     `Wow, you both have equally long names, ${hacker1.length} characters!`
   );
 }
+console.log();
 
 // Iteration 3: Loops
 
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -1,6 +1,125 @@
...
+  } else {
+    result += hacker1[i].toUpperCase() + " ";
+  }
+}
+console.log("3.1 solution with for loop: ", result);
+
+//  Exercice 3.1 done with methods
+console.log(
+  "3.1 solution with methods: ",
+  hacker1.split("").join(" ").toUpperCase()
```
TA-Remote, line 47: cool you're trying different approaches 💪

js/index.js
```diff
@@ -1,6 +1,125 @@
...
+    } else {
+      if (shortestName.length !== longestName.length) {
+        message = "The navigator's name goes first.";
+      } else {
+        message = "The names are the same.";
+      }
+    }
+  }
+}
+console.log("3.3 solution with for loop: ", message);
```
TA-Remote, line 111: In the first approach for it 3.3 , you are repeating a lot of code. You can loop only once, over the length of longest word .

And add additional check, for example:
```
hacker1[i] < hacker2[i] || hacker1[i] === undefined
```
// second condition will check if the letter of the driver is lower (goes first) or doesn't exist.

This way if there's no more letter at that index -> you have the answer, shorter word will go first

TA-Remote (commented):
Excellent Paul, keep it up!

PR #2066, opened 2022-06-15
Julie Bardiani, lab-javascript-basic-algorithms

Nika-Julie:


TA-Remote:
Good Job Julie !

js/index.js
```diff
@@ -1,6 +1,51 @@
...
+else if (hacker1.length === hacker2.length) {
+
+    console.log(`Wow, you both have equally long names, ${stringshacker1} characters!`)
+}
+
+
 
 // Iteration 3: Loops
+
+for (let i= 0 ; i < hacker1.length ; i++) {
```
TA-Remote, line 35: Here , we want to see a word with capital letters and with space between : "M A X" , not seprate letters , you need to declare a varible before loop, and add every letter in that :
let capitalName=""
for (let i= 0 ; i < hacker1.length ; i++) {
capitalName= capitalName+ hacker1[i]+" " 
  }
    console.log(capitalName.toUpperCase());

js/index.js
```diff
@@ -1,6 +1,51 @@
...
+    console.log(hacker1[i]);
+  }
+  
+  for (let i= hacker1.length -1; i>=0 ; i--) {
+    console.log(hacker1[i]);
+  }
```
TA-Remote, line 41: you need to have a word here again, like previous iteration ;)

PR #2069, opened 2022-06-15
Zubair Malek pull request

zubairmalek:
completed all solutions, that was fun but will definitely not be doing it again :)


TA-Remote:
```
completed all solutions, that was fun but will definitely not be doing it again :)
```

What won't you do again ;D

PR #2095, opened 2022-07-06
[FTLIS072022] Luisa Majid

marilu777:


js/index.js
```diff
@@ -1,6 +1,36 @@
...
+// Iteration 3
+let capitalName = "";
+let lowercaseName = "";
+
+for(let i = 0; i< hacker1.length; i++) {
+  console.log(hacker1[i]);
+}
+for(let i = hacker2.length-1; i>= 0; i--) {
+  console.log(hacker2[i]);
+}
```
webLisbonTA, line 28: Here it is not working as expected because when you console.log() inside of a loop it will print each character at a time in a new line, meaning that you would need to store the name in a variable and then console.log it like this: 
```
let newName = ""
   newName += hacker1[i].toUpperCase() + " ";
}
console.log(newName)
```
So, here on top we are creating a variable and we are adding each letter from the hacker1 name one by one with a space in between. Meaning that when we console.log it will all be in the same line as expected

webLisbonTA (commented):
Good job on the the lab!

PR #2097, opened 2022-07-06
done - Adriana Leitao and Philip Trachtenberg

AdrianaLeitao:


webLisbonTA:
Great job on completing the lab including the first bonus!! You were just missing a variable with the Lorem Ipsum but as soon as I added it, it worked fine, good job!

PR #2183, opened 2022-09-07
[RMT 5SEP] - Joao Mingote & Matthew Schröder

joaomingote:


js/index.js
```diff
@@ -1,6 +1,70 @@
...
+    console.log(driver)
+}
+
+// Iteration 3.2
+
+let navigator = '';
+for (let i = hacker2.length - 1; i >= 0; i--) {
+    navigator += hacker2[i];
+
+    console.log(navigator)
```
TA-Remote, line 38: If the `console.log` was outside the loop it will just print on time and the final result ^^

js/index.js
```diff
@@ -1,6 +1,70 @@
...
+
+  Aliquam cursus augue at massa pulvinar semper. Vestibulum placerat tristique euismod. Proin non elementum arcu. Fusce cursus sodales ante, sit amet pulvinar erat fringilla vitae. Vestibulum nec nisi congue libero maximus mollis. Aliquam facilisis in ligula eget ornare. Praesent laoreet enim at purus commodo, ut porta mi porta. Duis efficitur justo ipsum, nec hendrerit metus condimentum nec. Quisque pulvinar lacinia arcu id dignissim. Sed quis accumsan nisl, sed dapibus ex. Aliquam dignissim dui id risus molestie dictum.
+  
+  In ultricies consequat velit, viverra tempus leo porttitor id. Integer sagittis vel tortor facilisis cursus. Nulla vitae magna enim. Integer non urna enim. Praesent eget ligula iaculis, accumsan felis quis, dapibus mauris. Ut finibus lectus ut tempor pellentesque. Sed eleifend enim id mi aliquam tristique. Cras maximus ullamcorper ex et rutrum. Suspendisse gravida semper suscipit. Cras ultricies odio vel diam sagittis viverra. In ultrices id lacus id dapibus.'
```
TA-Remote, line 61: Here I was getting an error, maybe because GitHub broke the lines, but one way to avoid it is using the back ticks  => `

TA-Remote (commented):
Great work

PR #2187, opened 2022-09-07
[RMT-5SEP] Mog Rouhi, Volodymyr Khodyrev

Mog-Rouhi:


TA-Remote (commented):
Amazing work!!

In javaScript in general we use camelCase, the snake_case will work, but it's not expected when others are reading your code :)

Mog-Rouhi:
> Amazing work!!
> 
> In javaScript in general we use camelCase, the snake_case will work, but it's not expected when others are reading your code :)

Mog-Rouhi:
Thank you so much for the feedback!! :)

Mog-Rouhi:
Thank you so much for the feedback!! :)

PR #2190, opened 2022-09-07
[RMT-5SEP] Farzi, Jose

Njthon:


js/index.js
```diff
@@ -1,6 +1,49 @@
...
+    'Wow, you both have equally long names, ' + hacker1.length + ' characters!'
+  );
+}
 
 // Iteration 3: Loops
+
+console.log(hacker1.toUpperCase().split(''));
+
+console.log(hacker2.split('').reverse().join(''));
+
```
TA-Remote, line 31: The idea was to use loops on that iteration, you get almost all the correct should print a string and not an Array

js/index.js
```diff
@@ -1,6 +1,49 @@
...
+/* if ('hacker2'.localeCompare('hacker1') === 1) {
+    console.log("The driver's name goes first.")
+  }else if ('hacker1'.localeCompare('hacker2') === -1){
+    console.log("Yo, the navigator goes first definitely.")
+      }else {
+        console.log("What?! You both have the same name?")
+                 }*/
+
+let arr = [hacker1, hacker2];
+arr.sort();
```
TA-Remote, line 41: Here is the same thing as above, pay attention that not always `.sort()` behaviors as you think it would, see the examples => https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

PR #2204, opened 2022-09-15
done. I might try the bonus questions later if I have time. Not sure how to use terminal or console on VS.

ConjecturesandRefutations:
I have no idea how to use the console on Visual Studio, nor the terminal. I always use Git Bash for the terminal. I had to copy and paste everything onto CodePen in order to check the console log output.


Paulfresnel:
Hey Alfie, You can run the command **node script.js** (and replacing 'script.js' with your actual .js filename) to execute your javascript code directly from Visual Studio Code. Hope I could help

PR #2261, opened 2022-10-19
Jakob Heintze javascript basics done

Jheintze:


PRO-BROS:
I also faced a similar issue on my [site](https://prhocoding.com/) but thankfully I hire a developer which resolve that's issue.

PR #2527, opened 2023-02-22
lab 3 with eric

andrehcarvalho6:
we find this js exercises way funnier than the ones from css

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,73 @@
 // Iteration 1: Names and Input
-
+let hacker1 = 'andre';
+console.log('driver one is ' + hacker1);
+let hacker2 = 'eric';
+console.log('the navigator is ' + hacker2);
 
 // Iteration 2: Conditionals
+let driver = hacker1.length;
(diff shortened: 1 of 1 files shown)


TA-Remote:
Well done!!!
I see you used split-reverse-join to solve iterations 3.1 and 3.2, nice! 
Of course there are different solutions, one of them would it be go with a for loop( just sharing😉 )
[Three Ways to Reverse a String in JavaScript](https://www.freecodecamp.org/news/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb/)
Quick feedback on bonus 4:
Instead of two nested "if statement" you could also use an logical "and" operator --> &&
`  if (longText[i] === `e` &&  longText[i + 1] === `t`) { 
      count++;
    }`

Ale 🙂

andrehcarvalho6:
Thank you Alessandra, just send it to Eric 😀

TA-Remote ***@***.***> escreveu no dia quinta, 23/02/2023
à(s) 11:55:

PR #2528, opened 2023-02-22
Chris & Alex

Webster312:
Worked pretty well with Alex

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,32 @@
-// Iteration 1: Names and Input
+const hacker1 = "Chris"
+const hacker2 = "Alex"
+
+console.log("The drivers name is " + hacker1 )
+console.log("The navigators name is " + hacker2)
+
+if (hacker1.length > hacker2.length){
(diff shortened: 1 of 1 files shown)


TA-Remote:
Well Done!! 
Happy to hear you enjoyed working together!
Quick feedback about iteration 3.3: When you want to compare string and/or order by alphabetical order, we should keep in mind the the computer doesn’t follow the same rules of a dictionary. So lower then “<“ and bigger then “>” don’t work properly. 
[Lexicographic order - Wikipedia](https://en.wikipedia.org/wiki/Lexicographic_order)
To do so we use a method .localeCompare()
[String.prototype.localeCompare() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)
Good practice is also compare the two strings .lowerCase() In your code, for example if you change “Alex” with “alex” you’ll see a different result😉.

Ale 🙂

PR #2529, opened 2023-02-22
[RMT FT 20022023] Cristian Palao

Dookietrukie:
Hi, as I mentioned today in class, I had to do this lab alone because I had to leave early. Anyway, I LOVED IT. I enjoy very much doing JS exercises. 

On the other hand, I had some issues on the last bonus exercise, the palindrome one. It "works" but it has some bugs with some palindroms and it makes to program a bit buggy. For example, if you enter "amor roma" it works fine, but if you try "taco cat", which is also a palindrome, it says that it is not. I know that it is because the way the program works divides the string in half and there's a blank space in the middle that makes that one of the parts of the string is longer than the other, but I couldn't fix it. I will work a way around that ASAP. 

Anyway, nice lab ❤️ Looking forward for the next one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,12 @@ console.log(`The driver's name is ${hacker1}`);
 console.log(`The navigator's name is ${hacker2}`)
 
 // Iteration 2: Conditionals
-
+if (hacker1.length > hacker2.length) {
+    console.log(`The driver has the longest name, it has ${hacker1.length} characters!`)
+} else if (hacker1.length === hacker2.length) {
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+} else {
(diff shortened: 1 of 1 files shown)


TA-Remote:
Well done!!!
Happy you like it🤩
Very clean code and nice solutions! good that you're also commenting and keeping track of the logic.
The palindroms iteration wasn't the easiest to compleate, the right one for a super bonus!
As you can see in the lab solution a possible way to 'clean' the phrase is go for a loop  and keep anly the letters... 
`let cleanPhrase = ""
for (let i = 0; i < phraseToCheck.length; i++) {
  // only add characters between "A" and "z"
  if ((phraseToCheck[i] >= 'A' && phraseToCheck[i] <= 'z')) {
    cleanPhrase = cleanPhrase + phraseToCheck[i]
  }
}`
Ale 🙂

PR #2530, opened 2023-02-22
[FTRMT022023] Amina Mohamed Khedr - Cassandre Lavallard

amina247:
it was so much better than yesterday's lab
we enjoyed it so much

Cassandre was the driver for the first Part and I was the Driver for the Bonus.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,17 @@
 // Iteration 1: Names and Input
 
-
 // Iteration 2: Conditionals
-
 
 // Iteration 3: Loops
 
-console.log(test)
+const lastIndex = word.length - 1;
+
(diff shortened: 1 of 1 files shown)


TA-Remote:
Well Done!! 
Sure Javascript is better then flexbox🤩
Quick feedback about iteration 3.1: 
- Print the characters of the driver's name, separated by space, and in capital letters, i.e., "J O H N"
Beside using .toUpperCase(), as you did, we need to add an empty space between each letters. 
To do so we need to loop through the driver’s name… there are different way to get the same result, hope you’ll find the time to give it a try😉

Ale 🙂

PR #2534, opened 2023-02-22
[rmt ft 20 feb 2023]Nihat Algan - Peng Chen

nihatalgan:
We worked well and enjoyed it. Happy to cooperate each other.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,45 @@
 // Iteration 1: Names and Input
 
-
 // Iteration 2: Conditionals
 
-
 // Iteration 3: Loops
+
+console.log("i am ready");
+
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Nihat and Peng! Quick feedback on your lab.
Congrats on completing all of the iterations and pushing to the bonus, great work!
One small improvement would be to check how many words are in a paragraph. You could use the **split()** method instead which takes a pattern as a parameter and returns an array, in this case we could use the **split()** like: `longText.split(" ")` which would return an array with all of the words, since we passed a string with a space meaning that we want to separate each word of the provided string.
Other than that, keep up the good work!!

PR #2535, opened 2023-02-22
[WD-RTM-FT-Feb2023] Tuna and Mattie 

TunaErkmen:
We have done all iterations and one bonus!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,76 @@
 // Iteration 1: Names and Input
+const hacker1 = "Tuna";
+console.log(`The driver's name is ${hacker1}`);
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Tuna and Mattie! Quick feedback on your lab:
Great work on going after the bonus! Really clean work.
One small improvement would be for the 'et' counter, you could use the **.toLowerCase()** method to make sure all of the 'et' words are going to be counted.
Other than that, keep up the good work!

TunaErkmen:
Thank you for your feedback! It's very essential

PR #2537, opened 2023-02-22
[WD-RTM-FT-Feb2023] Paulius and Pim

pmockevicius:
We worked together on Lab and separated for bonus section. This homework was nice, not so demanding, fun to complete. Thank you

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,109 @@
-// Iteration 1: Names and Input
+// Iteration 1
+
+// 1.1
+const hacker1 = "Paulius"
+// 1.2
+console.log("The driver name is " + hacker1)
+// 1.3
+const hacker2 = "Pim"
+// 1.4
+console.log(`The navigator's name is ${hacker2}`)
 
 
-// Iteration 2: Conditionals
+// Iteration 2
+
+// 2.1
+if (hacker1.length == hacker2.length) {
+    console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`)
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Pim and Paulius! Quick feedback on your lab:
Great work on completing all of the iterations + bonuses! Clean work and props to the regex.
Only small improvement would be for the word count, as you are counting it by spaces it's missing one word, so instead you could use the `split()` method, that takes a pattern as an argument and returns an array of strings corresponding to the pattern. In this case we could do: 
`longText.split(" ").length`
So we would have an array with each word as an element of the array, since we passed a string with a space for the parameter of `split()`, meaning that we want to separate each word of the string.
Other than that, keep up the great work!

pmockevicius:
Thank you for the feedback Daniel !Much appreciatedOn Feb 24, 2023, at 4:28 AM, Daniel Skonetzky ***@***.***> wrote:﻿
Hey Pim and Paulius! Quick feedback on your lab:
Great work on completing all of the iterations + bonuses! Clean work and props to the regex.
Only small improvement would be for the word count, as you are counting it by spaces it's missing one word, so instead you could use the split() method, that takes a pattern as an argument and returns an array of strings corresponding to the pattern. In this case we could do:
longText.split(" ").length
So we would have an array with each word as an element of the array, since we passed a string with a space for the parameter of split(), meaning that we want to separate each word of the string.
Other than that, keep up the great work!

PR #2539, opened 2023-02-22
[FTRMT022023] Marin-Senem

valyy151:
We did not use the traditional methods, you said we should be resourceful and we were resourceful. :D


ykztenoks:
Hey Senem and Marin! Quick feedback on your lab:
Congrats on completing all of the iterations! Really clean work.
Small improvements would be for the 'et' counter, you could use the .toLowerCase() method to make sure all of the 'et' words are going to be counted. 
Other than that, keep up the good work!

PR #2540, opened 2023-02-22
{FTRMT022023] CherryLeung + David Correia

yi38leung:


TA-Remote:
well done!!
Small feedback:
- when we initialise variables ( string, number, and boolean) that we are going to manipulate or modify, is always better use      “let”—> let hacker1 = ‘Bob' 
The same for the long lorem ipsum text! 😉

Keep coding🙌
Ale

PR #2575, opened 2023-03-22
Done! Luis.Ramirez RMT WD-PT MAR23. First class lab iterations #1,2,3... but pending Bonuses

LuieTech:
This is my first PR! your feedback is very appreciated. I am still working on the Bonus challenges #1 and #2.
Luis.Ramirez RMT WD-PT MAR23

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,53 @@
 // Iteration 1: Names and Input
 
+let hacker1 = "Marcelo"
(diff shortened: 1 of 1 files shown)


index.js
LuieTech, line 1: Here I am submitting the first Bonus challenge from the day 1 Lab.  Please let me know if it's ok to use .split().length cause this is the only way i found to complete this task. Please let me know, Thank you

PR #2612, opened 2023-04-10
[LIS-WDFT-APR23] Margarida Pereira

mapmargaridapereira:


index.js
```diff
@@ -1,7 +1,46 @@
...
+
 
 // Iteration 3: Loops
+
+
+let hacker1Name = "";
+for (let i = 0; i <= hacker1.length - 1; i++) {
+ hacker1Name = hacker1Name + " " + hacker1[i];
+}
+ console.log (hacker1Name.toUpperCase());
```
joaopdg, line 30: You should use the `.toUpperCase()` method inside the loop, otherwise you are not saving the string with the upper cased letters, you are just console logging it:

```
let hacker1Name = "";

  for (let i = 0; i < hacker1.length - 1; i++) {
   hacker1Name = hacker1Name + " " + hacker1[i].toUpperCase()
  }

   console.log (hacker1Name);
```

Also, since index `[0]` starts counting from 0 and `.length` starts counting from 1, inside the for loop condition you should check if `i < hacker1.length` instead of `<=` otherwise the loop will run one more lap than what it was suppose to and you will get `undefined` on the last index

joaopdg (commented):
Great job on the lab Margarida, next time try to do the bonus!

PR #2622, opened 2023-04-15
JuanPablo.Maurer, Luis.Guillén WDMX-PT Apr 2023

churrundo:


churrundo:
I just committed an edit on the README file. Iteration 3 is titled "Loops", and that caused me and my partner to try using a loop over and over to solve the exercise. When we asked the instructor for help he said it just needed an if-else statement to evaluate a .localeCompare() method between the variables. Basically the inclusion of that exercise in that third section felt like a red herring that wasted us a lot of time. Looking at the previous iteration (Conditionals) I saw that there's only one exercise, so I think that last exercise on section 3 works best as a second exercise on section 2.

PR #2635, opened 2023-05-10
Nishtha Agarwal

nishag-26:


index.js
```diff
@@ -1,7 +1,30 @@
...
+if (i>j)
+{
+  console.log(`The driver has the longest name, it has ${i} characters.`);
+}else if(j>i) {
+  console.log(`It seems that the navigator has the longest name, it has ${j} characters.`);
+}
+  else if(i=j){
+  console.log(`Wow, you both have equally long names,${j} characters!.`);
+  }
+  for( k = 0; k < hacker1.length ; k++ ) {
```
TA-Remote, line 16: Even inside the loops use the `let` keyword, so this variable has block scope instead of being a global one

index.js
```diff
@@ -1,7 +1,30 @@
...
+console.log(`The navigator's Name is ${hacker2}`);
+let i=hacker1.length;
+let j=hacker2.length;
+if (i>j)
+{
+  console.log(`The driver has the longest name, it has ${i} characters.`);
+}else if(j>i) {
+  console.log(`It seems that the navigator has the longest name, it has ${j} characters.`);
+}
+  else if(i=j){
```
TA-Remote, line 13: A single equals means assignment in javascript, so the correct thing would be `==` or `===` (preferably triple equals)

index.js
```diff
@@ -1,7 +1,30 @@
...
+  console.log(`Wow, you both have equally long names,${j} characters!.`);
+  }
+  for( k = 0; k < hacker1.length ; k++ ) {
+    const strOriginal = hacker1[k];
+    const strUpperCase = hacker1[k].toUpperCase();
 
+  console.log(strUpperCase);
```
TA-Remote, line 20: Here you possible would want the `console.log(strUpperCase)` and the variable declaration outside of the loop, so you could print the string as a whole instead of inside of the `for loop`

TA-Remote (commented):
- Be mindfull of the indentation
- Careful using variables as i, j, k as those would be really hard to remember later or if you need to reuse them somewhere else this can become complicated to remember to which piece of information they're holding
- Keep up the work and practicing your skills

PR #2637, opened 2023-05-10
[FTRMT052023] Fernando / Pierre

fndoRamos:
1. 2. 3. Iterations done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,50 @@
 // Iteration 1: Names and Input
 
+let hacker1 = "Pierre";
+let hacker2 = "Fernando";
+
+console.log(`The driver's name is ${hacker2}`);
+console.log(`The navigator's name is ${hacker1}`);
 
 // Iteration 2: Conditionals
 
+if (hacker2.length > hacker1.length) {
+  console.log(
+    `The driver has the longest name, it has ${hacker2.length} characters.`
+  );
+} else if (hacker2.length < hacker1.length) {
+  console.log(`It seems that the navigator has the longest name, it has ${hacker1.length} characters.`);
+} else {
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
 
 // Iteration 3: Loops
+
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,7 +1,199 @@
...
+} else if (hacker1.localeCompare(hacker2) === 1) {
+  console.log("The driver's name goes first.");
+} else {
+  console.log("What?! You both have the same name?");
+}
+
+//Bonus
+
+//1-
+const longtext = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nunc a sagittis augue. Duis id urna aliquam erat iaculis accumsan. Sed tempus at dui a dictum. In non eros venenatis, tincidunt justo ut, elementum sapien. Nullam orci libero, pulvinar at dignissim auctor, rhoncus vel felis. Duis nisi mi, mattis tempus nulla a, varius ornare dolor. Cras ipsum dui, blandit eu imperdiet nec, dictum ut nisl. Aliquam euismod laoreet augue vel tempor. Praesent elementum enim gravida eros consectetur vehicula. Sed egestas ac magna et mollis. Pellentesque nulla urna, tempor vel orci sed, fringilla pretium nulla. Fusce nisi neque, ultrices non auctor vel, blandit quis magna. Donec tempor neque ac erat euismod interdum.
```
TA-Remote, line 59: Here we possibly want to use camelCase so it would be `longText` for example

index.js
```diff
@@ -1,7 +1,199 @@
...
+}
+
+count = count + 1;
+
+console.log(count, "the number of word in the variable longtext.");
+
+
+// 1.2 Count "Et".
+
+let count1 = 0;
```
TA-Remote, line 85: And we usually want to avoid using those kind of names `count1` maybe it could be `etCount` for example

TA-Remote (commented):
- Good work
- Good organization and identation
- Aside from the 2 variables that I mentioned, good naming

PR #2641, opened 2023-05-10
[FTRMT052023] Greg Tildis & Viktor Nestorov

gtildis:
Iterations 1,2,3: Done


TA-Remote (commented):
- Good work
- Good organization and indentation
- Good names in general
- If you have the time also trey the bonus iterations so you get more practice

PR #2642, opened 2023-05-10
[FTRMT052023] Linh-Carolin

linhvnde:


linhvnde:
will do bonus on my own pace

TA-Remote:
Hello @linhvnde and  @carolingoerner,
well done with your first javascript lab🙌
very clean and readable code, going step by step!!!
I guess you commented your solution because it was complaining when you were working on it. when we initialise a variable using const, we can not use the same name again. But when you declared 
const hacker1 = ‘Bob’ 
const hacker2 = ‘Alice’  ones , you can use them in all your iterations
Keep it up with your great job!😉!
Alessandra

PR #2645, opened 2023-05-10
[FTRMT052023] Maria Friedemann and Djamel Pettes

ProgrammingDj:
Iteration 1-3, Done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,56 @@
-// Iteration 1: Names and Input
+// Iteration 1: names and input
+const hacker1 = 'Maria';
+const hacker2 = 'Djamel';
 
+const driver = hacker1;
+const navigator = hacker2;
+
+console.log(`The driver's name is ${hacker1}`);
+
+console.log(`The navigator's name is ${hacker2}`);
 
 // Iteration 2: Conditionals
 
+if (driver.length > navigator.length) {
+  console.log(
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hello @ProgrammingDj and @[softcake1988](https://github.com/softcake1988)]
Well done on your first javascript lab🙌
I found your solution for it 3.3 very interesting, one of the best part of becoming a developer is that there are different way to solve the same problem🤩
.sort() is an array method, we still haven't seen in class yet (and I'm sure you know how it works), you'll have soon the opportunity to practice it more. For now I'd suggest to focus more on  string methods, they will come usefull very often in other labs and projects. 
Onother method we use to compare 2 strings and their alphabetical order is [.localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)
Give it a look😉
keep it up with great job!
Alessandra

ProgrammingDj:
Hello Alessandra,

yes we were debating if local compare or not, but we understood it would be
for two variables, the sort solution comes more handy for more variables,
that is why we choose that route. But yes you are right we should stick to
strings. Thank you for your feedback.

Kind regards

Djamel Pettes

TA-Remote ***@***.***> schrieb am Do. 11. Mai 2023 um 13:00:

PR #2646, opened 2023-05-10
[FTRMT052023] Filipe Ferreira and Diogo Martins

filipemgf:


index.js
```diff
@@ -1,7 +1,154 @@
...
 
 // Iteration 3: Loops
+
+let capitalsWithSpace = "";
+
+for (let i = 0; i < hacker1.length; i++) {
+	capitalsWithSpace += hacker1[i].toUpperCase() + " ";
+}
+
+console.log(capitalsWithSpace);
```
TA-Remote, line 32: Remember to use one keyword to declare your variables `let` or  `const`

index.js
```diff
@@ -1,7 +1,154 @@
...
+let countWords = 0;
+
+for (let i = 0; i < textArray.length; i++) {
+	countWords += 1;
+}
+
+console.log(countWords); */
+
+let countWords = 1;
+for (i = 0; i < longText.length; i++) {
```
TA-Remote, line 70: Remember to use the `let` keyword, even inside the `for`, otherwise this would create a global variable and in general we want to avoid that

index.js
```diff
@@ -1,7 +1,154 @@
...
+	reverseNavigatorName += hacker2[i];
+}
+
+console.log(reverseNavigatorName);
+
+console.log(capitalsWithSpace.localeCompare(reverseNavigatorName));
+
+if (capitalsWithSpace.localeCompare(reverseNavigatorName) > 0) {
+	console.log("Yo, the navigator goes first, definitely.");
+} else if (capitalsWithSpace.localeCompare(reverseNavigatorName) < 0) {
```
TA-Remote, line 46: Here maybe you could have used the `hacker1` or `hacker2` directly

TA-Remote (commented):
- Good work
- Good organization and identation of the code
- Good naming

PR #2647, opened 2023-05-10
[FTRMT052023] marySueMeissner-eirikKlebo 

MarySueMe:
We worked on this assignment together.


index.js
```diff
@@ -1,7 +1,124 @@
...
+    count++;
+  }
+}
+
+console.log(`The Latin word "et" appears ${count} times.`);
+
+//Bonus 2 (Don´t know how to remove space in between)
+
+let phraseToCheck = "taco cat";
+let phraseToCheckTrimmed = phraseToCheck.trim();
```
TA-Remote, line 112: The .trim() removes the white space from start and end, that's why it didn't solve your problem here.

index.js
```diff
@@ -1,7 +1,124 @@
...
+} else {
+  console.log(`What?! You both have the same name?`);
+}
+
+//Bonus 1
+
+function getWordCount(str) {
+  return str.split(" ").filter(function (n) {
+    return n != "";
+  }).length;
```
TA-Remote, line 89: That function gets the job done, but could you rewrite that using only a `for loop`? You could use a similar logic in the palindrome

TA-Remote (commented):
- Good work and nice to see that for some iterations you had more than one solution

PR #2651, opened 2023-05-10
[FTRMT052023] Marco

mk3346:
done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,38 @@
 // Iteration 1: Names and Input
-
+let hacker1 = "Marco";
+console.log(`The driver's name is ${hacker1}.`);
+let hacker2 = "Marco";
+console.log(`The navigator's name is ${hacker2}.`);
 
 // Iteration 2: Conditionals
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,7 +1,38 @@
...
+  console.log(`Wow, you both have equally long names, ${hacker1.length} characters!`);
+}
 
 // Iteration 3: Loops
+let hacker1UpperCase = hacker1.toUpperCase();
+let hacker1Separate = [];
+
+for (let i = 0; i < hacker1UpperCase.length; i++) {
+  hacker1Separate.push(hacker1UpperCase[i]);
+  console.log(hacker1Separate[i]);
```
TA-Remote, line 22: Here `hacker1Separate` probably should be a string and the `console.log()` possibly should be outside the `for loop` so you could print the whole string with spaces

TA-Remote (commented):
- Good work
- Good naming
- Good indentation / organization of the code

PR #2654, opened 2023-05-11
lab done

JCouderc87:
I did this alone as Marco did not join me. 
I really struggled with the last one, I did not know how to do this and Daniel had to explain it to me, but still remain very confused.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,30 @@
 // Iteration 1: Names and Input
 
+var hacker1 = { firstName: 'Jeremy', lastName: 'Couderc' };
+var message = 'The drivers name is ' + hacker1.firstName + ' ' + hacker1.lastName + '!!';
+console.log(message);
 
 // Iteration 2: Conditionals
 
+const message = `The driver has the longest name, it has XX characters`;
+
+console.log(message.length);
+
+const  XX = '53';
+const message2 = `The driver has the longest name, it has ${XX} characters`;
+console.log(message2);
 
 // Iteration 3: Loops
+
+const driver = "Jeremy";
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,7 +1,30 @@
 // Iteration 1: Names and Input
 
+var hacker1 = { firstName: 'Jeremy', lastName: 'Couderc' };
+var message = 'The drivers name is ' + hacker1.firstName + ' ' + hacker1.lastName + '!!';
```
TA-Remote, line 4: We usually avoid `var` to define our variables now, because they can escape scope and that can be troublesome for some applications

index.js
```diff
@@ -1,7 +1,30 @@
...
+console.log(message);
 
 // Iteration 2: Conditionals
 
+const message = `The driver has the longest name, it has XX characters`;
```
TA-Remote, line 9: Because you already have a message previously we wouldn't try to redeclare it again

index.js
```diff
@@ -1,7 +1,30 @@
...
+
+console.log(message.length);
+
+const  XX = '53';
+const message2 = `The driver has the longest name, it has ${XX} characters`;
```
TA-Remote, line 14: Here we possibly could have used `const message2 = `The driver has the longest name, it has ${hacker1.firstName.length} characters`;` to get the length of that string

TA-Remote (commented):
- Possible the reason for some strange parts is that you were producing the code in an Online IDE and pasting the code, but I'll comment just for you to be aware of that

- Keep practicing and the coding will get easier with time. Good job

PR #2705, opened 2023-07-10
[LIS-FT-JULY23] maureen_treacy

maureen-treacy:


index.js
```diff
@@ -1,7 +1,82 @@
...
+    if (twoChars === "et") {
+        numChars += 1
+    }
+}
+
+console.log(numChars)
+
+// Bonus 2
+
+let phraseToCheck = "amor roma"
```
brunorocha20, line 70: Here, you could try a different approach to this iteration. The way you did it, is just going to work if the sentence is exactly the same backward. It doesn't consider sentences that may be palindromes but have different structures. 
Here's another solution for this exercise:
```
let stringToCheck = 'A man, a plan, a canal, Panama!';
stringToCheck = stringToCheck.replace(/[ ,!.]/g, '').toLowerCase();
let reversedString = stringToCheck.split('').reverse().join('');
if (stringToCheck === reversedString) {
  console.log('String is a palindrome');
} else {
  console.log('Not a palindrome');
}

```

brunorocha20 (commented):
Good job on the lab, check the comments I left you for a better understanding.

PR #2711, opened 2023-07-10
[LIS-FT-JULY23] mariana_silva

maryanasilva:


index.js
```diff
@@ -1,7 +1,39 @@
...
+if (hacker1.length > hacker2.length) {
+  console.log("The driver has the longest name, it has 7 characters.");
+} 
+else {
+  console.log;
+}
 
 // Iteration 3: Loops
+
+const characters = hacker1.toUpperCase().split("").join(" ");
```
brunorocha20, line 21: Mariana, this is a correct solution for the exercise that we propose, however, you should've use a loop to reach the answer.
Here's another solution using a `for` loop:
```
  if (hacker1[i].localeCompare(hacker2[i]) === -1) {
    break;
  } else if (hacker1[i].localeCompare(hacker2[i]) === 1) {
    console.log("Yo, the navigator's name goes first definitely");
    break;
  } else if (i === hacker1.length - 1) {
    console.log('What?! You both have the same name?');
  } else if (hacker1[i].localeCompare(hacker2[i]) === 0) {
    continue;
  }
}
```

brunorocha20 (commented):
Good job on the lab, if you have time, try doing the bonus iterations as they are a good way to practice.

PR #2714, opened 2023-07-10
[LIS-FT-JULY23] Bruno_Cunha

brcunha3000:


brcunha3000:
I've done the labs again after reading the instructions
"IMPORTANT: If you use Google to help you to find a solution to this iteration, you might run into some advanced solutions that use string or array methods (such as join(), reverse(), etc.). However, we want you to apply your current knowledge and try to come up with a solution by just using the for loop and if-else statements with some break and continue."
Should i commit it again?

brunorocha20 (commented):
Nice job on the lab, Bruno.

PR #2792, opened 2023-10-11
[LIS-FT-OCT23] marina_balp

marifawellness:


webLisbonTA:
Hi Marina,
Great job on your lab!
On the first loop, the point was to have your name printed on the same line. Example: M A R I N A

One way of doing that would be:

let capitalDriversName = ''
for (let i=0; i<hacker1.length; i++){
capitalDriversName += hacker1[i].toUpperCase();
if(i < hacker1.length - 1){
capitalDriversName += ' ' // here you will add a space between each letter that is coming from the loop, except for the last one.
}
}

console.log(capitalDriversName)

Overall, good job.

PR #2795, opened 2023-10-11
[LIS-FT-OCT23] David_Lucca

davidvlucca:


jacobviv (commented):
Good job Mariana and David

index.js
```diff
@@ -1,7 +1,75 @@
 // Iteration 1: Names and Input
 
+const hacker1 = "David Lucca";
+console.log("The driver's name is "+hacker1);
```
jacobviv, line 4: Better this way:
```javascript
 console.log(`The driver's name is ${hacker1} `);
``` 
Same as you have done in the next iteration

index.js
```diff
@@ -1,7 +1,75 @@
...
+else if ((`${hacker1}`).localeCompare(`${hacker2}`) === 1){
+    console.log("Yo, the navigator goes first, definitely.");
+}   
+else {
+    console.log("What?! You both have the same name?");
+}
+
+
+
+if(hacker1 > hacker2){
```
jacobviv, line 51: This option should be working changing the order:
```javascript
if(hacker1 < hacker2){
    console.log("The driver's name goes first.");
}
else if (hacker1 > hacker2){
    console.log("Yo, the navigator goes first, definitely.");
}
else {
}
```

index.js
```diff
@@ -1,7 +1,75 @@
...
+else if (hacker1 < hacker2){
+    console.log("Yo, the navigator goes first, definitely.");
+}
+else {
+    console.log("What?! You both have the same name?");
+}
+
+
+/*
+Alternative option
```
jacobviv, line 63: And this alternative option also need to flip the options:
```javascript
if (comparison === -1) {
  console.log("The driver's name goes first.");
} else if (comparison === 1) {
  console.log("Yo, the navigator goes first, definitely.");
} else {
  console.log("What?! You both have the same name.");
}
```

PR #2820, opened 2023-10-18
[FTRMT2023] Marat Pomeranets

pomerama:
We did some nice talking & programming together with Vidisha, but decided to deliver our separate code:)

Question: is there a more efficient way to check for palindrome? My version works, but it looks ugly:)


pomerama:
cool, thanks!

PR #2837, opened 2023-10-18
CAMILA WDFTRMOCT23 

camilaflorez-dev:


camilaflorez-dev:
Hii, I have to practice a lot more so that it is more automatic to understand what is the right way to get the results, Gustavo was very helpful, especially explaining .split to solve one of the exercises. i didn't that well with the 2nd bonus ...

gusfonte99:
Yeah this one was pretty simple, I think.
 In bonus 2 I thought of using .split() to get rid of everything but the letters.
Then, using a for loop, add each letter in reverse order to a new string.
And finally an if statement to compare both strings (original and reversed), if everything worked, supposedly it would return true and then print "This string is a palindrome".

PS: sorry to submit after the start of class, i just wanted to make sure i was doing it right (as per what i wrote in slack) :)

TA-Remote:
Hello @camilaflorez-dev & @gusfonte99 
well done with your solution🙌
I'm very happy to hear you worked together helping each other, that’s the developers spirit!!
About bonus 2: it can be a way to go, keep in mind you might want to ‘clean up’ the initial phase from all the characters that are not letters… just a hint😉  
but you’ll have a lot of opportunities to practice different patters and methods  
Keep it up💪  
Ale

